PR #293, opened 2026-02-04
feat(security): add data flow security with agent hook integration (Spec 027)

Dumbris:
## Summary

- Add data flow security system that detects and prevents data exfiltration by tracking how data moves between internal tools (Read, databases) and external tools (WebFetch, Slack)
- Support two operating modes: **proxy-only** (universal, works with any MCP agent) and **full mode** (with agent hook integration for intercepting agent-internal tool calls like Read, Write, Bash)
- Implement complete hook CLI (`mcpproxy hook install/uninstall/status/evaluate`) for Claude Code integration

### Key Components

- **Classifier** (`internal/security/flow/classifier.go`): Categorizes tools/servers as internal, external, hybrid, or unknown using name heuristics and config overrides
- **Content Hasher** (`internal/security/flow/hasher.go`): SHA256 per-field extraction from JSON for content flow matching without storing raw data
- **Flow Tracker** (`internal/security/flow/tracker.go`): Session-scoped origin recording and flow edge detection with configurable policies
- **Policy Evaluator** (`internal/security/flow/policy.go`): Configurable actions (allow/warn/ask/deny) with graceful degradation in proxy-only mode (ask→warn)
- **Session Correlator** (`internal/security/flow/correlator.go`): Links agent hook sessions to MCP proxy sessions via argument hash matching
- **Hook CLI** (`cmd/mcpproxy/hook_cmd.go`): Install/uninstall/status/evaluate commands for Claude Code agent hooks
- **REST API** (`internal/httpapi/hooks.go`): `POST /api/v1/hooks/evaluate` endpoint for hook event processing
- **Activity Logging**: New `hook_evaluation` and `flow_summary` activity types with full metadata
- **Web UI Nudge**: Dashboard hint encouraging hook installation when running in proxy-only mode
- **Diagnostics**: Security coverage recommendation in `mcpproxy doctor` output
- **OpenAPI**: Updated swagger.yaml with new endpoint, activity filters, and schemas

### Architecture

```
Agent (Claude Code)
  ├── Internal tools (Read, Write, Bash) ──→ Hook CLI ──→ POST /api/v1/hooks/evaluate
  └── MCP tools (call_tool_read/write) ──→ MCP Proxy ──→ Upstream servers
                                              ↕
                                    Flow Tracker (session-scoped)
                                    Content hash matching
                                    Policy evaluation
```

## Test plan

- [x] 68/70 TDD tasks complete (2 benchmarks deferred to pre-merge)
[...]

PR #299, opened 2026-02-07
fix(security): resolve race condition and resource leaks in flow tracker

tjsingleton:
## Summary

Fixes 3 issues found during code review of #293 (Spec 027 data flow security):

- **Race condition**: `CheckFlow()` writes to session fields under read lock — changed `RLock` to `Lock`
- **Goroutine leak**: `FlowService.Stop()` never called in `Runtime.Close()` — added shutdown call
- **Missing callback**: `SetExpiryCallback` never registered — flow session summaries now emit to activity log

## Changes

**Production fixes** (`2611a32`):
- `internal/security/flow/tracker.go`: `RLock` → `Lock` in `CheckFlow` (writes to `LastActivity`, `ToolsUsed`, `Flows` require write lock)
- `internal/runtime/runtime.go`: Added `flowService.Stop()` in `Close()` before supervisor shutdown
- `internal/runtime/runtime.go`: Registered `SetExpiryCallback` after Runtime construction to emit `EventTypeActivityFlowSummary`

**Regression tests** (`ab95259`):
- `TestFlowTracker_CheckFlow_ConcurrentRace`: 15+ goroutines exercising concurrent `CheckFlow` + `RecordOrigin`
- `TestFlowService_Stop_Idempotent`: triple-stop safety
- `TestFlowService_ExpiryCallback_EmitsSummary`: verifies callback fires with correct `FlowSummary`

## Test plan

- [x] `go test -race ./internal/security/flow/...` — all 34 tests pass
- [x] Race regression test confirms fix (would fail with `RLock`)
- [x] Code review verified: no deadlock risk, correct shutdown order, non-blocking callback

> Stacked on #293

### DIFF
--- a/internal/runtime/runtime.go
+++ b/internal/runtime/runtime.go
@@ -241,6 +241,24 @@ func New(cfg *config.Config, cfgPath string, logger *zap.Logger) (*Runtime, erro
 		phaseMachine: newPhaseMachine(PhaseInitializing),
 	}
 
+	// Register flow expiry callback now that rt is available (Spec 027)
+	if rt.flowService != nil {
+		rt.flowService.SetExpiryCallback(func(summary *flow.FlowSummary) {
+			rt.EmitActivityFlowSummary(
+				summary.SessionID,
+				summary.CoverageMode,
+				summary.DurationMinutes,
+				summary.TotalOrigins,
+				summary.TotalFlows,
+				summary.FlowTypeDistribution,
+				summary.RiskLevelDistribution,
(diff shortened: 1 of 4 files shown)


Dumbris:
Thank you for the thorough code review and the fixes, @tjsingleton! Catching the race condition in `CheckFlow()` (read lock vs write lock), the goroutine leak from missing `FlowService.Stop()`, and the unregistered expiry callback — these are exactly the kind of subtle issues that matter in production. Great contribution!

PR #367, opened 2026-04-08
fix(release): auto-update Homebrew cask on release

tjsingleton:
Closes #366

## Problem

The `update-homebrew` job only updates `Formula/mcpproxy.rb` (CLI tarballs). `Casks/mcpproxy.rb` (macOS DMG / tray app) was never touched, leaving it stranded at v0.20.2 while the formula reached v0.23.1.

## Fix

Two new steps added to the existing `update-homebrew` job:

1. **Download installer DMGs and calculate SHA256 for cask** — downloads `mcpproxy-{version}-darwin-{arm64,amd64}-installer.dmg` with the same retry logic used for tarballs, exports `SHA256_CASK_ARM64` and `SHA256_CASK_AMD64` env vars.

2. **Update Homebrew cask** — patches `Casks/mcpproxy.rb` in-place: bumps `version` and replaces both `sha256` values. Skips gracefully if the cask file doesn't exist.

The existing **Commit and push changes** step is updated to also stage `Casks/mcpproxy.rb`.

## Notes

- No new secrets or permissions needed — uses the existing `HOMEBREW_TAP_TOKEN`
- DMG assets are produced by the `release` job before `update-homebrew` runs, so they should be available
- The sed approach for updating the two sha256 values is order-dependent (arm64 first, amd64 second) — matches the current cask structure

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1348,12 +1348,63 @@ jobs:
           echo "Formula created successfully:"
           cat Formula/mcpproxy.rb
 
+      - name: Download installer DMGs and calculate SHA256 for cask
+        run: |
+          VERSION=${GITHUB_REF#refs/tags/v}
+          BASE_URL="https://github.com/${{ github.repository }}/releases/download/v${VERSION}"
+
+          for ARCH in arm64 amd64; do
+            DMG="mcpproxy-${VERSION}-darwin-${ARCH}-installer.dmg"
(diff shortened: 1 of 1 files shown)


Dumbris:
Thanks for the contribution, @tjsingleton! Great catch on the stranded cask — the fix is clean, reuses the existing retry pattern, and the sed order matches the cask file layout. Merged.  🙏

PR #422, opened 2026-04-26
fix(scanner): drop MCP-MC-001 false positive on synthetic tools.json

Dumbris:
## Summary

- Filters findings that fire on the synthetic `tools.json` metadata file mcpproxy writes for the AI scanner, when the rule is the obfuscation/encoded-payload rule (MCP-MC-001 and look-alike variants).
- Tool-poisoning, prompt-injection, and other tool-semantics rules **still** fire on `tools.json` — that's why we export it.

## Why

`tools.json` is generated by `Service.exportToolDefinitions` via `json.MarshalIndent`. Go's JSON encoder emits non-ASCII chars in tool descriptions as legitimate `\uXXXX` escape sequences. The mcp-ai-scanner's MCP-MC-001 rule treats those as "Source code contains obfuscated or encoded payloads that may hide malicious behavior" and surfaces a phantom **HIGH** "Malicious Code" finding for any server whose tool descriptions contain a single non-ASCII character.

The user hit this on `obsidian-pilot`: scanner reported `Risk Score: 25/100` and a HIGH finding at `tools.json:85` with no actual malicious content — the location was just the JSON serialization of a tool description.

This is independent of #421 (which fixes the upstream Docker-extraction bug that *increased* exposure to this false positive by making `tools.json` the only file in the scan dir). Both should land.

## Approach

Smallest-correct fix is post-aggregation filter:

```go
func filterSyntheticToolDefinitionFindings(findings []ScanFinding) []ScanFinding
```

Drops a finding when **both**:
- `Location` resolves to `tools.json` (any path depth, any `:line` suffix), AND
- The rule is the obfuscation rule — `RuleID == "MCP-MC-001"`, or `Category` contains `malicious` and Description/Title contains "obfuscat" / "encoded payload" (defensive coverage for variant scanners).

Tool-poisoning rules (`MCP-TPA-*`, `Category=tool_poisoning`) and other rules survive on `tools.json` unchanged.

## Test plan

- [x] `TestFilterSyntheticToolsJSONObfuscation` — 7 cases (drop/keep matrix: absolute paths, real source files, tool-poisoning on tools.json kept, tools.jsonl near-miss, empty location).
- [x] `TestLocationIsSyntheticToolsJSON` — path-suffix stripping handles different scanner formats.
- [x] `TestAggregateReportsDropsObfuscationOnToolsJSON` — integration test verifying wiring + that legitimate findings survive.
- [x] `go test ./internal/security/scanner/ -count=1` passes.
- [x] `golangci-lint run ./internal/security/scanner/...` clean.


Dumbris:
Closing without merging — the user pushed back on this approach (https://github.com/smart-mcp-proxy/mcpproxy-go/pull/422 hard-coded `MCP-MC-001` in `engine.go`). mcpproxy needs to stay universal: scanner-specific rule IDs and false-positive workarounds belong in the scanner repo, not in the host.

The fix has been moved to `mcp-ai-scanner` itself (`~/repos/mcpproxy-scanner-claude-code`). The scanner is being patched so `MCP-MC-001` no longer matches `\uXXXX` escape sequences inside data/config files (`.json`, `.yaml`, `.toml`, `.ini`, `.cfg`) where Unicode escapes are legitimate encoding rather than obfuscation. After the patched image rebuilds and is published, users will get the fix automatically via the standard scanner image pull on next scan.

Branch `fix/scanner-tools-json-obfuscation-fp` retained in case the universal mechanism (synthetic-file manifest convention scanners can opt into) ever becomes worth pursuing. For now: scanner repo is the source of truth.

PR #425, opened 2026-04-26
security(headers): redact sensitive header values in upstream_servers list and /api/v1/servers

electrolobzik:
# Pull Request

## Description
Both the upstream_servers MCP tool's list operation and the REST GET /api/v1/servers endpoint were returning the per-server Headers map verbatim — so any caller of those endpoints (an MCP
  agent inside a sandbox, a local tool with the API key, the Web UI, the macOS tray) could read Authorization: Bearer <token>, X-API-Key, session cookies, and other secrets out of another
  upstream's config.

  Concretely observed: a sandboxed MCP agent calling mcpproxy:upstream_servers operation=list got back a registered GitHub server's Authorization: Bearer ghp_… PAT in plaintext.

### What changes

  - internal/oauth/logging.go — new RedactStringHeaders(map[string]string) map[string]string that mirrors the existing RedactHeaders(http.Header) helper but takes the form actually used in
  cfg.Headers. Same sensitiveHeaders allowlist, same ***REDACTED*** marker, returns a copy (input not mutated).
  - internal/server/mcp.go (handleListUpstreams) — redacts the headers field on every server entry before serialising the tool result.
  - internal/httpapi/server.go (handleGetServers) — new private redactServerHeaders helper applied in both the management-service code path and the legacy fallback before writeSuccess.
  - internal/config/config.go — new RevealSecretHeaders bool field (reveal_secret_headers JSON / reveal-secret-headers mapstructure key, default false). Setting it to true opts out of redaction
   for users who genuinely need raw values surfaced (e.g. a custom UI that round-trips full PUT replaces).

  Sensitive header names redacted (case-insensitive, list lives in oauth.sensitiveHeaders): Authorization, X-API-Key, Cookie, Set-Cookie, X-Access-Token, X-Refresh-Token, X-Auth-Token,
  Proxy-Authorization.

 ### What does NOT change

  PATCH / merge semantics. The underlying stored cfg.Headers is untouched — only the read-side response is redacted. So upstream_servers operation=patch continues to deep-merge by name; only
  the value displayed in subsequent reads is masked. Existing callers that omit unchanged fields keep working exactly as before.

 ### Migration

  None. Existing config files keep working; reveal_secret_headers defaults to false. To opt out: {"reveal_secret_headers": true} in mcp_config.json.

## Testing

Unit tests added in internal/oauth/logging_test.go (TestRedactStringHeaders, 8 sub-tests): redacts Bearer Authorization, redacts X-Api-Key, redacts Cookie/Set-Cookie, preserves non-sensitive
  headers (Content-Type), case-insensitive header name matching, nil input returns nil, empty input returns empty map, input map is not mutated.

  E2E tests updated in internal/server/e2e_test.go:
  - TestE2E_PatchDeepMergesEnvAndHeaders — now asserts Authorization == "***REDACTED***" (key preserved by merge, value masked in API output) while X-Custom and X-New-Header retain their actual
   values, validating both that merge semantics still work and that redaction is scoped to the sensitive-name list.
  - TestE2E_MultipleEnableDisablePreservesConfig — same pattern: Authorization key survives 5 enable/disable toggles, with the value redacted in read-back.

  Local test runs (clean):
  ok  github.com/smart-mcp-proxy/mcpproxy-go/internal/oauth
  ok  github.com/smart-mcp-proxy/mcpproxy-go/internal/server    (incl. the two E2E tests above)
  ok  github.com/smart-mcp-proxy/mcpproxy-go/internal/httpapi
  ok  github.com/smart-mcp-proxy/mcpproxy-go/internal/config
  ok  github.com/smart-mcp-proxy/mcpproxy-go/internal/contracts
  go build ./... clean.

  Manual verification: end-to-end call against the running proxy now returns ***REDACTED*** for the GitHub server's Authorization header (previously returned the live PAT).

- [x] All existing tests pass

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -163,6 +163,15 @@ type Config struct {
 	// Security scanner settings (Spec 039)
 	Security *SecurityConfig `json:"security,omitempty" mapstructure:"security"`
 
+	// RevealSecretHeaders, when true, disables the redaction of sensitive
+	// header values (Authorization, X-API-Key, Cookie, …) in responses from
+	// the `upstream_servers` MCP tool and the `/api/v1/servers` REST API.
+	// Default false — sensitive header values are surfaced as
+	// `***REDACTED***` so an MCP agent cannot read Bearer tokens / API keys
+	// out of another upstream's config. Set to true if a downstream tool
+	// needs the raw values (e.g. a UI that round-trips PUT replaces).
+	RevealSecretHeaders bool `json:"reveal_secret_headers,omitempty" mapstructure:"reveal-secret-headers"`
+
 	// Server edition multi-user configuration (only meaningful with -tags server)
 	Teams *TeamsConfig `json:"teams,omitempty" mapstructure:"teams" swaggerignore:"true"`
 }
--- a/internal/httpapi/server.go
+++ b/internal/httpapi/server.go
@@ -22,6 +22,7 @@ import (
 	"github.com/smart-mcp-proxy/mcpproxy-go/internal/contracts"
 	"github.com/smart-mcp-proxy/mcpproxy-go/internal/logs"
 	"github.com/smart-mcp-proxy/mcpproxy-go/internal/management"
+	"github.com/smart-mcp-proxy/mcpproxy-go/internal/oauth"
 	"github.com/smart-mcp-proxy/mcpproxy-go/internal/observability"
 	"github.com/smart-mcp-proxy/mcpproxy-go/internal/reqcontext"
 	internalRuntime "github.com/smart-mcp-proxy/mcpproxy-go/internal/runtime"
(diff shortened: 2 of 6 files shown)


Dumbris (approved):
Thank you for the contribution! Clean security fix — sensible default, opt-out provided, well-tested. 🙏

Dumbris:
Thanks for the security fix and the thorough writeup — approved! 🙏

CI is red on **Verify OpenAPI Artifacts**: the new `RevealSecretHeaders` field on `Config` gets picked up by `swag`, so `oas/swagger.yaml` (+10 lines) and `oas/docs.go` (+1 line) need to be regenerated and committed. The check ran the regen and found uncommitted diffs.

Two ways to unblock:

**Option A — quickest:** enable *Allow edits and access to secrets by maintainers* in the PR sidebar, and we'll push the regen.

**Option B — push it yourself:**
```bash
make swagger
git add oas/swagger.yaml oas/docs.go
git commit -m "chore(oas): regenerate OpenAPI artifacts for reveal_secret_headers"
git push
```

Either way, once that lands and CI is green we'll merge. Thanks again!

Dumbris:
Merging now via admin override — bypassing the OAS check, will push the `make swagger` regen as a follow-up commit on main. Thanks again for the fix! 🙏

PR #427, opened 2026-04-26
fix(upstream/core): normalize nil CallTool arguments to empty object

electrolobzik:
# Pull Request

## Description
Some MCP servers reject null for params.arguments and expect an empty object {} instead. The most visible offender is BrowserMCP, which uses strict JSON-Schema validation and returns an error
   when called with no arguments.

  When Client.CallTool is invoked with args == nil (legitimate for tools that take no arguments), the resulting JSON-RPC request serialises params.arguments as null, which those servers reject.
   The fix normalises a nil map to map[string]interface{}{} immediately before assigning it to request.Params.Arguments, so the wire form becomes {}.

### What changes

  internal/upstream/core/client.go — three lines added in CallTool, before the existing request.Params.Arguments = args:

  if args == nil {
      args = map[string]interface{}{}
  }

  No behaviour change for non-nil inputs; servers that previously accepted null keep working.

## Testing

go build ./... clean. The change has been running locally for ~5 weeks against BrowserMCP without issue (BrowserMCP-style tools were unable to be called at all before; they all work after).

  I held off on adding a focused unit test because exercising Client.CallTool cleanly requires standing up a fake MCP transport; happy to add one if you'd like — let me know whether to extract a buildCallToolRequest helper for testability or to wire up a transport mock.

- [x] All existing tests pass

### DIFF
--- a/internal/upstream/core/client.go
+++ b/internal/upstream/core/client.go
@@ -348,6 +348,9 @@ func (c *Client) CallTool(ctx context.Context, toolName string, args map[string]
 
 	request := mcp.CallToolRequest{}
 	request.Params.Name = toolName
+	if args == nil {
+		args = map[string]interface{}{}
+	}
 	request.Params.Arguments = args
 
 	// Log to server-specific log


Dumbris (approved):
Thanks for the fix and for the context — BrowserMCP-style strict-schema servers definitely need this. The 3-line change is correct; merging as-is. 🙏

PR #428, opened 2026-04-26
chore(macos-tray): untrack swiftc-generated module cache

electrolobzik:
# Pull Request

## Description

The file `native/macos/MCPProxyUITest/Swift-CR65AB5HVZD2.swiftmodule` is a swiftc build cache (the `Swift-` prefix + hex hash + `.swiftmodule` extension is the cache-naming convention swiftc uses for ad-hoc single-file compilations). It gets rewritten on every Swift build of `MCPProxyUITest`, so it surfaces as a continuous "modified content" diff for anyone who has ever built the package — and the live-rewrite also propagates as a `<sha>-dirty` flag to the submodule pointer in any repo that submodules mcpproxy-go.

The per-directory `.gitignore` at `native/macos/MCPProxyUITest/.gitignore:2` already lists `*.swiftmodule`, but it was ineffective for this one file because the file got tracked before the rule was in place (or via `git add -f`). `git rm --cached` removes the index entry; the existing `.gitignore` rule then suppresses it cleanly going forward. Working tree file is untouched (it's regenerated on the next build anyway).

Same intent as the existing `native/macos/MCPProxy/.build/` rule in the root `.gitignore` — keep auto-generated swiftc artifacts out of the source tree.

## Testing
- [x] All existing tests pass


Dumbris (approved):
Thank you for pointing this out

PR #446, opened 2026-05-03
Fix intermittent missing enabled servers in tool retrieval

JasonLandbridge:
### Summary

This PR fixes an intermittent issue where tools from enabled/connected servers (notably `rider-official`) would sometimes not appear in `retrieve_tools` results, even though upstream `ListTools` calls succeeded.

The fix addresses two core causes:

1. **Concurrent `ListTools` contention** produced transient failures (`ListTools operation already in progress`) during discovery.
2. **Discovery race windows** could temporarily omit connected servers from global discovery output, causing visible tool-set shrink in search results.

---

## Root cause

During server churn (restart/reconnect, reactive discovery, background discovery), multiple discovery paths could call `ListTools` on the same managed client at the same time.

Previously:
- first call proceeded,
- second call failed fast with an “already in progress” error.

At the same time, global discovery accepted partial results from currently successful servers. If one connected server was omitted in that cycle, its tools were temporarily absent from indexed/searchable results.

This produced the observed “sometimes shows, sometimes missing” behavior.

---

### 1) Coalesced concurrent `ListTools` calls
**File:** `internal/upstream/managed/client.go`

- Reworked managed client `ListTools` concurrency behavior:
  - one caller performs upstream `ListTools`,
  - concurrent callers wait for the same in-flight operation,
  - all waiters receive the shared result/error.
- Eliminates fail-fast path that previously returned:
  - `ListTools operation already in progress for server ...`

**Impact:** prevents transient per-server tool discovery drops caused by local concurrency collisions.

---

### 2) Added last-good per-server snapshot fallback in global discovery
**Files:** `internal/runtime/runtime.go`, `internal/runtime/lifecycle.go`

- Added runtime cache of last known good tool snapshots per server.
- During `DiscoverAndIndexTools`:
  - fresh per-server snapshots are updated from current discovery results,
  - standard differential updates are applied,
  - for connected servers missing from the current discovery pass, the last-good snapshot is reused and re-applied.

**Impact:** avoids temporary index shrink when a connected server is missed in one discovery cycle.

---

### 3) Regression tests for ListTools coalescing
**File:** `internal/upstream/managed/listtools_coalescing_test.go`

Added tests:
- `TestListTools_CoalescesWaiters`
- `TestListTools_CoalescesWaitersError`

These verify that waiters receive shared success/error instead of being rejected while a list operation is already in progress.

---

## Verification

- `go test ./internal/upstream/managed ./internal/runtime` passed.
- Build check passed (`build_project` success).

---

## Scope and non-goals

- This PR focuses on correctness/stability of discovery and visibility.
- Does **not** change retrieval ranking semantics or hard limits (e.g., `retrieve_tools.limit` max 100).
- Does **not** introduce debounce/metadata enhancements (can be follow-up).

---

## Expected behavior after fix

- Enabled/connected servers should no longer disappear from tool retrieval due to in-process `ListTools` races.
- Transient discovery omissions should be masked by last-good snapshots, reducing inconsistent tool visibility across adjacent queries.

---

## Follow-ups (optional)

- Add explicit discovery health metadata to `retrieve_tools` responses (`fresh/degraded`, missing/stale servers).
- Add event debounce for server churn bursts to reduce redundant discovery passes.
- Add broader integration test simulating reconnect storms with repeated retrieval queries.

### DIFF
--- a/internal/runtime/lifecycle.go
+++ b/internal/runtime/lifecycle.go
@@ -302,13 +302,54 @@ func (r *Runtime) DiscoverAndIndexTools(ctx context.Context) error {
 		toolsByServer[tool.ServerName] = append(toolsByServer[tool.ServerName], tool)
 	}
 
-	// Apply differential update for each server
+	// Persist fresh snapshots for discovered servers
+	r.lastGoodToolsMu.Lock()
 	for serverName, serverTools := range toolsByServer {
+		cp := make([]*config.ToolMetadata, len(serverTools))
+		copy(cp, serverTools)
+		r.lastGoodTools[serverName] = cp
+	}
+	r.lastGoodToolsMu.Unlock()
+
+	// Apply differential update for each server with fallback to last-good snapshots
+	processedServers := make(map[string]struct{}, len(toolsByServer))
+	for serverName, serverTools := range toolsByServer {
+		processedServers[serverName] = struct{}{}
 		if err := r.applyDifferentialToolUpdate(ctx, serverName, serverTools); err != nil {
 			r.logger.Error("Failed to apply differential update for server",
 				zap.String("server", serverName),
 				zap.Error(err))
 			// Continue with other servers instead of failing completely
+		}
+	}
+
+	// For connected servers that were temporarily missing from discovery results,
+	// re-apply last-good snapshot to avoid transient index shrink.
+	for _, serverName := range r.upstreamManager.GetAllServerNames() {
+		if _, ok := processedServers[serverName]; ok {
+			continue
+		}
+
+		client, ok := r.upstreamManager.GetClient(serverName)
+		if !ok || client == nil || !client.IsConnected() {
+			continue
+		}
+
+		r.lastGoodToolsMu.RLock()
+		snapshot, hasSnapshot := r.lastGoodTools[serverName]
+		r.lastGoodToolsMu.RUnlock()
+		if !hasSnapshot || len(snapshot) == 0 {
+			continue
+		}
+
+		r.logger.Warn("Server missing from discovery result; reusing last-good tool snapshot",
+			zap.String("server", serverName),
+			zap.Int("snapshot_tools", len(snapshot)))
+
+		if err := r.applyDifferentialToolUpdate(ctx, serverName, snapshot); err != nil {
+			r.logger.Error("Failed to apply last-good snapshot for server",
+				zap.String("server", serverName),
+				zap.Error(err))
 		}
 	}
 
--- a/internal/runtime/lifecycle.go
+++ b/internal/runtime/lifecycle.go
@@ -302,12 +302,27 @@ func (r *Runtime) DiscoverAndIndexTools(ctx context.Context) error {
 		toolsByServer[tool.ServerName] = append(toolsByServer[tool.ServerName], tool)
 	}
 
-	// Persist fresh snapshots for discovered servers
+	// Snapshot the set of currently-known servers so we can prune entries for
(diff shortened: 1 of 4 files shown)


Dumbris:
Pushed [`0fdbc4d8`](https://github.com/JasonLandbridge/mcpproxy-go/commit/0fdbc4d8) addressing the review. Summary of changes:

### 1. Fixed nil-channel deadlock for non-ListTools leaders

`acquireListToolsContext` is still used by the health check (`client.go:817`) and the tool-count refresh (`client.go:1290`), but in the previous version of this PR it didn't allocate `listToolsWaitCh`. If either of those held the in-progress flag and a `ListTools` call arrived, the new waiter loop did:

```go
waitCh := mc.listToolsWaitCh   // nil
select {
case <-ctx.Done():
case <-waitCh:                 // nil channel — blocks forever
}
```

That would hang for the full 30s `ListTools` timeout on every coalesced call, strictly worse than the old fail-fast.

Fix: `acquireListToolsContext` always allocates the wait channel and clears the cached result on leader election; both non-coalescing call sites now call the new `publishListToolsResult` helper before `release()` so any coalesced `ListTools` waiter gets a real result instead of `nil`/stale data.

### 2. Consolidated leader bookkeeping

`ListTools` was duplicating the leader-acquire logic from `acquireListToolsContext`. It now goes through that helper for the leader path (`runListToolsAsLeader`) — single source of truth for `inProgress` / `waitCh` / `cancel` lifecycle. The waiter loop handles the genuine race where the previous holder releases between our failed acquire and the re-check (rare, but possible) by retrying once before parking on the channel.

### 3. Snapshot map no longer grows unboundedly

`DiscoverAndIndexTools` now prunes `r.lastGoodTools` entries for servers no longer in `GetAllServerNames()`. Previously a removed/renamed server's snapshot lingered for the lifetime of the process.

### 4. Demoted self-healing log to Info

`"Server missing from discovery result; reusing last-good tool snapshot"` is benign self-healing and can fire repeatedly during reconnect storms — switched from `Warn` to `Info`.

### 5. Real concurrency tests

The original two tests in `listtools_coalescing_test.go` only exercised the waiter-reads-cached-result branch (preset state, then call `ListTools`). They never proved the leader path or the channel close ordering. Added:

- `TestListTools_MultipleWaitersUnblockedByLeader` — real leader + 5 waiter goroutines, verifies all wake up and see the same published result.
- `TestListTools_HealthCheckLeaderDoesNotDeadlockWaiters` — explicit regression test for the nil-channel bug. Spins a `ListTools` waiter behind a health-check style leader (acquired via `acquireListToolsContext` + `publishListToolsResult`) with a 2s ctx timeout. Fails the suite with `"ListTools waiter did not wake up after leader released"` if coalescing is broken.
- `TestListTools_LeaderErrorPropagatedToWaiters` — error fan-out.
- `TestListTools_AcquireContextResetsCachedResult` — guarantees a new leader election can never return stale data from a previous round.

### Verification

```
go test ./internal/upstream/managed/ ./internal/runtime/ -race -count=1
ok      .../internal/upstream/managed   1.372s
ok      .../internal/runtime            90.053s
go build ./...   # clean
```

The `ToolMetadata` shallow-copy point from the review is left as-is (with a comment noting the immutability assumption); converting to a deep copy would touch a wider surface and the values aren't mutated post-discovery today.

Dumbris (approved):

Dumbris:
Thanks @JasonLandbridge! Merged. Solid root-cause work on the discovery race - the coalescing primitive is now the single source of truth for in-flight ListTools calls, which makes the next round of upstream-management changes a lot easier to reason about.

PR #447, opened 2026-05-04
Allow tools to be disabled and thus hidden from mcp discovery

### LINKED ISSUE #432: [Feature] Add ability to disable individual tools in MCP servers after approval
Hi there,

Once an MCP server has been added and its tools have been approved, there does not currently seem to be a way to go back and block or disable a specific tool.

Image

This would be useful when multiple MCP servers expose overlapping tools. For example, a user may want to keep both servers enabled, but force a specific tool call to come from only one of them by disabling the duplicate tool on the other server.

Would it be possible to add a per-tool toggle after approval?

Suggested behavior:

* Each approved tool can be enabled or disabled individually.
* Disabled tools are no longer exposed to MCP clients.
* Disabled tools are hidden or visually marked as disabled in the tool cache list.
* The setting persists across restarts.

This would make it easier to manage overlapping MCP servers without having to disable an entire server.

JasonLandbridge:
Closes #432
image
image

image

## Summary
- treat disabled tools as non-callable/non-discoverable via centralized callable checks
- filter disabled tools before retrieve_tools response construction
- return TOOL_BLOCKED with clear message for direct calls to disabled tools
- exclude disabled tools from server status health tool_count displays
- add regression tests for callable predicate, discovery filtering, blocked call error, and re-enable visibility

## Validation
- go test ./internal/server -run 'TestIsToolCallable_DisabledTool|TestRetrieveTools_ExcludesDisabledToolsPreRanking|TestCallBlockedTool_ReturnsToolBlocked|TestReenableTool_VisibleAndCallable|TestHandleListUpstreams_FilteredForAgent'

## Notes
- security/admin introspection remains available via explicit quarantine/security endpoints
- standard discovery and call planning no longer surfaces disabled tools


Dumbris:
Reviewed and pushed [`a8d01147`](https://github.com/JasonLandbridge/mcpproxy-go/commit/a8d01147) with three fixes plus the missing `make swagger` regeneration.

## Review summary

The core mechanism — gating tool exposure on `ToolApprovalRecord.Disabled` via a centralized `isToolCallable` predicate that's checked in both `handleRetrieveTools` and `handleCallToolVariant`/`handleCallTool` — is sound. Tests pass with `-race`. A few issues worth fixing before merge:

### 1. Startup regression in `handleListUpstreams` (fixed)

`getVisibleToolCount` reads from the supervisor's StateView snapshot, which is **empty during startup before the supervisor has hydrated**. The fallback to live `ListTools` was gated on `p.mainServer == nil` — i.e. only ever runs in tests/standalone — so in production a freshly-connected server with healthy tools would render as **"0 tools"** in the UI until the next reconciliation cycle.

Restored the live fallback in production too, but filtered through `isToolCallable` so disabled tools still don't inflate the count:

```go
toolCount = p.getVisibleToolCount(server.Name)
if toolCount == 0 {
    if tools, err := client.ListTools(context.Background()); err == nil {
        for _, tool := range tools {
            if p.isToolCallable(server.Name, tool.Name) {
                toolCount++
            }
        }
    }
}
```

### 2. Per-result `auth.AuthContextFromContext(ctx)` (fixed)

In the new `handleRetrieveTools` filter loop, the auth context lookup was inside the per-result `for` iteration:

```go
for _, result := range results {
    ...
    if authCtx := auth.AuthContextFromContext(ctx); authCtx != nil && !authCtx.IsAdmin() {
        if !authCtx.CanAccessServer(serverName) { continue }
    }
    ...
}
```

`context.Value` is a hash lookup per call, and the auth context is a per-request value — repeated work for every search result. Hoisted it (and the `IsAdmin` check) to a single resolution above the loop.

### 3. `TestRetrieveTools_ExcludesDisabledToolsPreRanking` silent-passed (fixed)

The previous form had a quiet escape hatch:

```go
toolsValue, exists := payload["tools"]
if !exists || toolsValue == nil {
    return  // ← silent pass
}
```

…and the indexed pair (`context7:resolve-library-id` disabled, `github:get_file_contents` enabled) didn't share keywords with the BM25 query (`context7 resolve library id`), so the only hit was the disabled tool. Once filtered out, the response had no `tools` field, and the test silently passed without verifying anything — including failing to catch a regression where the github tool also got accidentally filtered.

Rewrote the fixture so both indexed tools share the same description (`"library documentation lookup helper"`) and BM25 returns both; replaced the silent return with `require.True` and added a positive control:

```go
assert.Contains(t, names, "github:library-lookup",
    "enabled tool must survive ranking and filtering")
assert.NotContains(t, names, "context7:library-lookup",
    "disabled tool must be filtered out before ranking")
```

### 4. `make swagger` regeneration (fixed)

The new `blocked_count` field on `QuarantineStats` wasn't reflected in `oas/swagger.yaml`/`oas/docs.go`. The pre-push hook caught this — regenerated and included.

## Verification

```
go test ./internal/server -count=1 -timeout 10m   PASS (434s)
go test ./internal/runtime -count=1                PASS
go test ./internal/httpapi -count=1                PASS
go vet ./...                                       clean
```

## Other observations (not changed)

- **Frontend `router/index.ts`**: removal of `scrollBehavior` looks unrelated to the spec - worth either a separate PR or a sentence in the description so reviewers know why it's bundled.
- **`isToolCallable` BBolt cost**: two BBolt reads per result (`GetUpstreamServer` + `GetToolApproval`). Fine at current `retrieve_tools` `limit ≤ 100`, but if that ever loosens, batch-fetching disabled tool names per server up-front would scale better.
- **Stale `_test.go` fixture in `tool_quarantine_test.go`**: `ParamsJSON: \`\\"schema\\"\`` (escaped quotes inside a backtick string) parses as the literal text `\"schema\"`, not a valid JSON string. Hashes still match because both sides use the same literal, but the data isn't what the field is documented to hold. Cosmetic.
- **Disabling does not invalidate live MCP sessions**: an in-flight tool call started before the disable will still complete. If you want strict cut-over semantics, you'd need to thread the check into the per-call middleware. Probably fine as-is — disabling is administrative.

Dumbris (approved):

Dumbris:
Thanks @JasonLandbridge! Merged. The TOOL_BLOCKED + isToolCallable centralization is a clean pattern - we'll be able to reuse it for the per-agent visibility work coming next.

JasonLandbridge:
Thanks!

PR #453, opened 2026-05-12
fix(oauth): treat zero ExpiresAt as never-expires in HasPersistedToken

electrolobzik:
## Summary

Fix an OAuth re-auth loop that breaks any upstream whose authorization server omits `expires_in` from its `/token` response (observed concretely against Atlassian's MCP endpoint at `https://mcp.atlassian.com/v1/mcp`).

The authorization succeeds end-to-end — the callback is received, the token is exchanged, the record lands in BBolt — but on every subsequent reconnect the upstream emits:

```
core/connection_oauth.go | MCP initialize JSON-RPC call failed | error: transport error: failed to send request: no valid token available, authorization required
managed/client.go        | all authentication strategies failed, last error: OAuth authentication required for <server>
```

and drops back to the same defer-OAuth-for-tray path it would take if no token existed.

## Root cause

`HasPersistedToken` (`internal/oauth/config.go`) computed `isExpired = time.Now().After(record.ExpiresAt)`. When the auth server doesn't return `expires_in`, the oauth2 library leaves `ExpiresAt = time.Time{}` (the Go zero value), and `time.Now().After(time.Time{})` is always true. The persisted token therefore looked expired on every reconnect, `skipBrowserFlow` stayed false in all four `HasPersistedToken` callers in `internal/upstream/core/connection_oauth.go` (HTTP precheck/post-check, SSE precheck/post-check), and the upstream looped forever asking for re-auth even though a valid bearer token was sitting in storage.

The asymmetry was the bug: `HasValidToken` (config.go:299-304) and mcp-go's own `Token.IsExpired()` (`client/transport/oauth.go:82-86` in v0.44.0-beta.2) both already treat `IsZero()` as "never expires". `HasPersistedToken` was the lone outlier.

## Changes

- `HasPersistedToken`: short-circuit on `record.ExpiresAt.IsZero()` to match the rest of the stack.
- `PersistentTokenStore.GetToken`: stop logging "OAuth token has expired and needs refresh" for zero-expiry tokens, and stop computing the negative grace-period adjustment that path produced — the existing `adjustedExpiresAt := record.ExpiresAt` fallback already kept the zero, so the token is returned unchanged for the underlying `client.Token`.
- Three new unit tests against `HasPersistedToken` covering zero / future / past expiries, so the IsZero branch can't regress.

No migration needed — existing records that were silently broken just start working on the next reconnect.

## Test plan

- [x] `go test ./internal/oauth/` — 14.9 s, all green
- [x] `go test ./internal/upstream/...` — all green
- [x] New `TestHasPersistedToken_ZeroExpiresAt` reproduces the Atlassian scenario and passes
- [x] Existing `TestTokenStoreManager_HasValidToken_PersistentStore_NoExpiration` still passes (no regression in the symmetric helper)
- [x] Reproduced fix end-to-end against `https://mcp.atlassian.com/v1/mcp` with a patched binary — the upstream stays connected across restarts after a single OAuth login

### DIFF
--- a/internal/oauth/config.go
+++ b/internal/oauth/config.go
@@ -100,6 +100,13 @@ func (m *TokenStoreManager) HasTokenStore(serverName string) bool {
 
 // HasPersistedToken checks if a token exists in persistent storage (BBolt) for the server.
 // This is the preferred method to check for existing tokens as it reflects actual token availability.
+//
+// A zero ExpiresAt (time.Time{}) is treated as "no expiration" — some authorization servers
+// (e.g. Atlassian's MCP endpoint) issue access tokens without an `expires_in` field, which
+// the oauth2 library deserializes as the Go zero time. Returning isExpired=true for those
+// records would make the upstream connection loop forever asking for re-auth even though a
+// valid token is on disk. This mirrors HasValidToken (see below), which already treats
+// IsZero() as "never expires".
 func HasPersistedToken(serverName, serverURL string, boltStorage *storage.BoltDB) (hasToken bool, hasRefreshToken bool, isExpired bool) {
 	if boltStorage == nil {
 		return false, false, false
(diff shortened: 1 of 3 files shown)


Dumbris:
Thanks Roman! 🙏 Great catch on the asymmetry between `HasPersistedToken` and `HasValidToken` / mcp-go's own `IsZero()` semantics — that was a sneaky one, and the Atlassian repro made it concrete. The fix is tight, the three regression tests (zero / future / past expiry) cover exactly the right triangle, and the log-message switch in `PersistentTokenStore.GetToken` is a nice cleanup.

Merging.

Issue #438, opened 2026-04-28
[BUG] Stale cache issues


JasonLandbridge:
Hi there!

I keep running into what looks like stale cache/state issues in the UI.

## Example

When I approve all tools:

<img width="3229" height="1282" alt="Image" src="https://github.com/user-attachments/assets/9bb4c915-1520-422d-9d9f-f468066522a9" />

Then navigate back to the overview page, the UI still shows that there are unapproved tools:

<img width="3247" height="1303" alt="Image" src="https://github.com/user-attachments/assets/bd5e8fc3-079a-4c87-844b-ac1ce7d34e45" />

However, if I refresh the page manually, the overview updates correctly and shows that everything is approved.

## Expected behaviour

After approving all tools, the overview page should immediately reflect the updated approval state without requiring a full browser refresh.

## Actual behaviour

The action appears to succeed, but other parts of the UI still show the previous state until the page is refreshed.

## Why this is confusing

I have noticed similar stale-state behaviour in several areas of the UI. It makes it difficult to know whether an action actually succeeded or whether the displayed state is outdated.

This is especially confusing when navigating between pages, because the UI can show different information depending on whether the user refreshed the page or only used client-side navigation.

## Possible cause / suggestion

I do not know the Golang at all, but from a frontend perspective this looks like state invalidation or cache synchronization is missing after mutations.

In my own project, [Reaparr](https://github.com/Reaparr/Reaparr), the [frontend](https://github.com/Reaparr/Reaparr/tree/dev/src/AppHost/ClientApp) is based on Vue, Nuxt, RxJS, and [Pinia stores](https://github.com/Reaparr/Reaparr/tree/dev/src/AppHost/ClientApp/src/store). The approach I use there is:

- API data flows into Pinia stores
- Vue components only consume reactive store state / computed values
- Actions that mutate server-side state also update or refresh the relevant store state
- Other components automatically update because they depend on the same reactive source of truth

That kind of pattern helps avoid stale cache issues, because the API result from an action is propagated through the shared frontend state instead of each page keeping its own outdated copy.

For this project, maybe mutations like approving tools could either:

- update the shared frontend state directly after success, `toolsStore.ts`
- return the updated server/tool state and use that to update the UI
- have the overview page depend on computed properties from the `toolsStore.ts`

## Impact

This would make the UI feel much more reliable, especially for setups with many MCP servers and many tools, where manually refreshing after every action becomes a hassle.

[...]

PR #463, opened 2026-05-13
feat(webui): per-tool enable/disable + bulk Enable All/Disable All

electrolobzik:
## Summary

Closes the gap between the per-server enable/disable that already exists and the per-tool admin toggle UI that was wired only halfway. The storage-backed `ToolApprovalRecord.Disabled` mechanism + the runtime filtering in `isToolCallable` / `retrieve_tools` / `handleCallToolVariant` were all already in place — but three issues prevented users from actually reaching the toggle:

1. `Runtime.SetToolEnabled` returned an error when no approval record existed for the tool — i.e. for every tool when `quarantine_enabled` was `false` (the default) or `skip_quarantine` was set on the server. Now it synthesizes a baseline `approved` record on demand so the toggle works in every configuration.
2. The Web UI per-tool toggle button was gated on `approval_status === 'approved'`, so non-quarantined tools (the common case) never saw it.
3. There was no bulk path. Operators with chatty servers had to click each tool individually.

## What this adds

**Backend**
- `internal/runtime/tool_quarantine.go`: `SetAllToolsEnabled(serverName, enabled, updatedBy) (changed int, err error)`. Walks the StateView + index + storage union of known tools and toggles only those not already in the target state — so the audit trail and SSE traffic stay clean on repeated calls.
- `internal/server/server.go`: `SetAllToolsEnabled` controller wrapper.
- `internal/httpapi/server.go`: `POST /api/v1/servers/{id}/tools/enable_all` and `.../disable_all`. The existing `GET /api/v1/servers/{id}/tools` now also surfaces the per-tool `disabled` flag (new `contracts.Tool.Disabled` field) so the frontend can render the toggle without a second round-trip to the approvals endpoint.
- `internal/runtime/event_bus.go`: SSE `servers.changed` payload now enriches each embedded server with `Quarantine{Pending,Changed,Blocked}Count` (mirroring `httpapi.enrichServersWithQuarantineStats`). **This also fixes a pre-existing tool-quarantine staleness bug**: the Web UI's `mergeServers` treats incoming server data as authoritative and deletes absent fields, so on every SSE delivery the previously-fetched `Quarantine` block got wiped — manifesting as quarantine badges and the new "N disabled" pill going stale until manual page refresh.
- `oas/swagger.yaml` + `oas/docs.go` regenerated via `make swagger`. `make swagger-verify` clean.

**CLI**
- `internal/cliclient/client.go`: `SetToolEnabled` + `SetAllToolsEnabled` HTTP clients.
- `cmd/mcpproxy/upstream_cmd.go`: new `mcpproxy upstream tools` subcommand group with `enable <server> <tool>`, `disable <server> <tool>`, `enable-all <server>`, `disable-all <server>`. Mirrors the verbs of `mcpproxy upstream enable|disable --all` so the surface is consistent.

**Web UI** (`frontend/src/views/ServerDetail.vue`, `frontend/src/components/ServerCard.vue`)
- Bulk **Enable All** / **Disable All** buttons in the Tools tab header — each visible only when there is actually something for it to do (avoids button labels that promise a no-op).
- Per-tool toggle now also renders for tools without an approval record (the common case).
- Per-tool Enable/Disable colored button replaced with a daisyUI `toggle toggle-sm` switch, the same widget used for the server-level Enabled toggle on the Config tab — same affordance applied at different scopes.
- New `syncAfterToolToggle` helper refreshes the store-backed servers list, re-syncs the local server ref, and reloads tools + approvals after any per-tool or bulk toggle. Eliminates the lag where the in-page "N disabled" pill (and quarantine badges) stayed stale until the next SSE.
- User-visible wording "blocked" → "disabled" everywhere (Server Detail stat-desc and Server Card status row) so it matches the server-level Enabled/Disabled toggle and doesn't imply external blocking.
- `frontend/src/types/contracts.ts`: `Tool.disabled` + `Tool.approval_status` are now typed so consumers don't fall back to `any`.

**Tests**
- `TestSetToolEnabled_CreatesRecordWhenMissing` — the no-record path that was the root cause of (1).
- `TestSetAllToolsEnabled_DisablesAllKnownTools` — bulk runtime method, including the idempotent re-call.
- `TestHandleSetAllToolsEnabled_DisableAll` / `_EnableAll` / `_ControllerError` — handler-level coverage for the new routes.

## Commits

1. `feat(webui): per-tool enable/disable + bulk Enable All/Disable All` — the main feature.
2. `fix(cli): replace deprecated strings.Title in upstream tools commands` — golangci-lint staticcheck SA1019 (caught by CI).
3. `fix(ux): refresh server quarantine stats on SSE + per-tool sync + wording` — the SSE quarantine-stats parity, sync-after-toggle, wording change, and toggle-switch widget.
4. `fix(ux): more per-tool toggle polish — corner toggle, server-state sync` — minor UX follow-ups.
5. `refactor(ux): derive server from store (computed), drop snapshot ref` — systematic fix for the stale local-UI bug class on `ServerDetail.vue`.
6. `perf(runtime): build servers.changed payload lazily in coalescer drainer` (post-review follow-up — see Update below).
7. `fix(quarantine): split SetToolEnabled, gate synthesis on sentinel error` (post-review follow-up).
8. `fix(runtime): honour app-shutdown cancellation in coalescer drainer` (post-review follow-up).
9. `fix(sse): plumb SecurityScan via management.ListServers so REST + SSE share enrichment` (post-review follow-up).
10. `fix(ux): reload tools/approvals/logs on serverName navigation (ServerDetail.vue)` (post-review follow-up).

## Update (2026-05-13): post-review follow-up commits

Five additional commits land the architectural cleanup recommended in the project's `PR463-FOLLOWUP-PLAN.md`. They extend Spec 047 §B2's amortisation guarantee from publish-side to build-side, fix the bulk-emit storm a different way than the original sketch (no global suppression flag), close one Spec 032 rug-pull bypass, close the SecurityScan SSE-parity gap, and fix the deep-link tool/approvals/logs staleness on `/servers/foo` → `/servers/bar` navigation.

**6. `perf(runtime): build servers.changed payload lazily in coalescer drainer` (10527bd)**

Spec 047 §B2 coalesced `servers.changed` publishes to ≤1 per 50ms window, but the build still ran eagerly inside `emitServersChanged`: a `ListServers` call plus an N-row `ListToolApprovals` per server. Bulk operations that fired K rapid emits paid K×(1+N) BBolt ops despite the coalescer dropping K-1 publishes — Enable-All on a 30-server / 50-tool setup was ≈1,500 wasted BBolt reads per click.

The coalescer now stores a lightweight `(reason, extra)` marker instead of a fully-realised `Event`. The drainer materialises the SSE payload at flush time via `buildServersChangedPayload`. Every emit is essentially free; the build runs once per publish window regardless of how many submits land. The no-coalescer fallback in `emitServersChanged` still builds inline.

Regression test: `TestCoalescer_AmortisesBuildAcrossBurst` asserts 100 rapid emits → exactly 1 `ListServers` call.

**7. `fix(quarantine): split SetToolEnabled, gate synthesis on sentinel error` (6032ed8)**

Two related fixes:

- Extract `setToolEnabledNoEmit(...) (changed bool, err error)`. `SetAllToolsEnabled` now calls the no-emit core directly in its loop and emits exactly one trailing `servers.changed` event (reason: `tools_enabled` / `tools_disabled`), mirroring the existing `ApproveAllTools` pattern in the same file. Lifts the "already in desired state" pre-check into the no-emit core so single-toggle calls also avoid no-op BBolt writes + SSE emits.
- Synthesis-on-not-found now requires `errors.Is(err, storage.ErrToolApprovalNotFound)`. `GetToolApproval` returns the new sentinel (wrapped). Any other error (decode failure, closed DB, mmap remap during compaction) is propagated to the caller. Without this gate a transient I/O error could silently demote a `pending`/`changed` record to `approved` — the Spec 032 rug-pull bypass risk.

New tests: `TestSetToolEnabled_PreservesExistingPendingStatus`, `TestSetAllToolsEnabled_EmitsOncePerBulk`, `TestSetAllToolsEnabled_NoEventOnNoOp`.

**8. `fix(runtime): honour app-shutdown cancellation in coalescer drainer` (5572611)**

Regression caught by code-review while landing #6. `flushNow` now does the build, which calls `ListServers` with a `context.WithTimeout(context.Background(), 2s)` — rooted in `Background`, not `appCtx`. After `Runtime.Close()` fires `appCancel`, the drainer goroutine could sit on the detached 2-second timer before exiting (`Close` doesn't wait on the drainer, so the practical impact is a leaked goroutine doing wasted work, not user-visible shutdown latency — but it silently broke the original Spec 047 promise that `flushNow` was O(1)).

Threads the parent ctx through: the coalescer stores it in `start()`, `flushNow` uses it as the parent for `buildServersChangedPayload`'s `WithTimeout`. Regression test: `TestBuildServersChangedPayload_HonoursCancelledParentCtx` asserts the call returns in < 200ms (vs. 2s) when the parent ctx is already cancelled.

**9. `fix(sse): plumb SecurityScan via management.ListServers so REST + SSE share enrichment` (818a06e)**

The servers.changed SSE embed enriched `Quarantine` (post-PR-463) but not `SecurityScan`. The REST `GET /api/v1/servers` handler enriched both via inline `securityController.GetScanSummary` calls. mergeServers on the Web UI deletes absent keys on incoming payloads — so every tool toggle silently stripped `security_scan` from every server in the store, blanking risk-score badges until the next REST refresh. Same bug class as the pre-existing quarantine-stats staleness this PR already fixes for Quarantine.

Fix: plumb scan summaries through `management.ListServers` (which BOTH the REST handler and `runtime.buildServersChangedPayload` call) so the two paths can't drift out of parity. New `management.SecurityScanEnricher` interface + `SetScanSummaryEnricher` setter; `internal/server` wires a `scanSummaryEnricherAdapter` that bridges `scanner.Service.GetScanSummary` (scanner-internal type) to `contracts.SecurityScanSummary` (the wire shape REST and SSE consumers already expect). The duplicate inline enrichment in `internal/httpapi/server.go` is dropped.

New tests: `TestListServers_populates_SecurityScan_via_enricher` + `TestListServers_nil_enricher_is_a_no-op` (management/service_test.go), `TestEmitServersChanged_PayloadPreservesSecurityScan` (runtime/event_bus_payload_test.go).

**10. `fix(ux): reload tools/approvals/logs on serverName navigation (ServerDetail.vue)` (d584b16)**

Vue Router 4 reuses the `ServerDetail.vue` component instance across `/servers/foo` → `/servers/bar` (same route, different param). The `server` computed correctly retargets via the Pinia store, but the local data refs (`serverTools`, `toolApprovals`, `serverLogs`, scan*) stayed populated with the previous server's data. Navigating between server detail pages showed server B's name + stats with server A's tool list until a manual refresh — looked like a data-corruption bug.

Adds `watch(() => props.serverName)` that resets every per-server local ref and re-runs `loadServerDetails`. UI-state refs (`activeTab`, `logTail`, `toolSearch`) are intentionally not reset so the user's tab/scope choice persists across navigation.

Race protection: introduces a `loadGeneration` counter bumped on every `loadServerDetails` entry. The three parallel fetches now go through internal `_loadToolsWithGen` / `_loadToolApprovalsWithGen` / `_loadLogsWithGen` helpers that capture the generation at entry and only commit results if it hasn't advanced — protects the foo→bar→foo case where foo's response arrives AFTER bar's load already started. Public no-arg `loadTools` / `loadToolApprovals` / `loadLogs` wrappers preserve existing call sites (template `@click` handlers, the connected/enabled watch, post-action refreshes) unchanged.

## Test plan

- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [x] `go test ./internal/runtime/ ./internal/httpapi/ ./internal/management/ ./internal/storage/ -count=1` — all pass
- [x] `npx vue-tsc --noEmit` — clean
- [x] `make swagger-verify` — OAS is committed and matches generated output
- [x] New: `TestCoalescer_AmortisesBuildAcrossBurst` — bulk amortisation regression test
- [x] New: `TestSetAllToolsEnabled_EmitsOncePerBulk` / `_NoEventOnNoOp` — bulk emit-once regression tests
- [x] New: `TestSetToolEnabled_PreservesExistingPendingStatus` — sentinel-error rug-pull regression test
- [x] New: `TestBuildServersChangedPayload_HonoursCancelledParentCtx` — shutdown-drain regression test
- [x] New: `TestEmitServersChanged_PayloadPreservesSecurityScan` — SSE SecurityScan parity test
- [x] New: `TestListServers_populates_SecurityScan_via_enricher` — management enricher wiring test
- [x] Manual: bring up the Web UI on a server with quarantine **disabled** globally and confirm the per-tool toggle now appears
- [x] Manual: disable a tool, then re-enable it, and confirm the "N disabled" pill clears immediately without a page reload
- [ ] Manual: navigate Server Detail → Servers list → back and confirm tool-quarantine / disabled-tool counts stay accurate (no stale state)
- [ ] Manual: navigate `/servers/foo` → `/servers/bar` and confirm bar's tool list / approvals / logs replace foo's (no stale data, no flash of A-then-B)
- [ ] Manual: confirm the SecurityScan risk-score badge stays accurate after any tool toggle (used to blank until the next REST refresh)
- [ ] Manual: click `Disable All` → confirm every tool flips, refresh page, confirm state persists
- [ ] Manual: `mcpproxy upstream tools disable-all <server>` from the CLI then re-list to confirm


electrolobzik (commented):
### Code review

Found 2 issues (focused on regressions + security per the request).

1. **SSE `servers.changed` embed only enriches `Quarantine`, not `SecurityScan` — every SSE delivery wipes `security_scan` from the Web UI store.** The new enrichment at [`internal/runtime/event_bus.go`](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/928ac93845713ee76bafd452119ebd189d1eae36/internal/runtime/event_bus.go#L184-L205) mirrors the REST handler's `enrichServersWithQuarantineStats`, but the REST handler [also enriches each server with `SecurityScan`](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/928ac93845713ee76bafd452119ebd189d1eae36/internal/httpapi/server.go#L1040-L1063) (Spec 039). The Web UI's `mergeServers` was specifically designed (issue #438 / PR #443) to treat incoming as authoritative and delete absent keys — the [explicit warning comment](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/928ac93845713ee76bafd452119ebd189d1eae36/frontend/src/stores/servers.ts#L53-L68) calls out exactly this pattern. Result: every `servers.changed` event (enable/disable, restart, quarantine change, tool toggle — many per session) silently strips `security_scan` from each store server, blanking the risk-score badge / scan summary until the next manual REST refresh. Same bug class the PR itself fixes for `Quarantine`. Fix: also enrich `SecurityScan` in `emitServersChanged`, or drop the embed for fields the SSE path can't mirror.

2. **Bulk `SetAllToolsEnabled` emits `servers.changed` per tool, each emit re-scans every server's BBolt approvals — partially reverses Spec 047's CPU win.** [`SetAllToolsEnabled`](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/928ac93845713ee76bafd452119ebd189d1eae36/internal/runtime/tool_quarantine.go#L685-L731) calls `SetToolEnabled` in a loop, and `SetToolEnabled` [emits `servers.changed` per invocation](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/928ac93845713ee76bafd452119ebd189d1eae36/internal/runtime/tool_quarantine.go#L666-L676). Each emit runs the new [`enrichServersWithQuarantineStats`](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/928ac93845713ee76bafd452119ebd189d1eae36/internal/runtime/event_bus.go#L213-L253) which does one `ListToolApprovals` per server. For K=50 tools × N=30 servers a single Disable-All produces ~1500 BBolt `View` calls. The Spec 047 coalescer collapses the *publish* but not the per-submit work — exactly the pattern eae45ef (#450, "cut idle CPU 92%") and #444 ("emit once per bulk call", see [`ApproveAllTools`'s single trailing emit](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/928ac93845713ee76bafd452119ebd189d1eae36/internal/runtime/tool_quarantine.go#L640-L646)) were designed against. Fix: emit `servers.changed` once at the end of the bulk loop (suppress per-tool emits when invoked from the bulk path), or cache approval counts in `StateView` and read from memory instead of BBolt.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Dumbris (approved):

Dumbris:
Thank you for this — really well-scoped and thoroughly tested work. Shipping per-tool enable/disable plus the bulk Enable All / Disable All actions closes a gap that's been bugging operators with chatty servers for a while, and the SSE quarantine-stats parity fix was a nice catch along the way.

Merging now. Thanks again, Roman.

PR #472, opened 2026-05-16
fix(generate-types): re-sync hardcoded TS output with contracts.ts + drift test

electrolobzik:
## Summary

The TypeScript-as-Go-string literal inside `cmd/generate-types/main.go` has drifted from the actual `frontend/src/types/contracts.ts` it claims to generate. Running `go run ./cmd/generate-types` (invoked by the `Makefile`'s `frontend-build` target on every `make build` / `make build-server`) silently reverts a chunk of `contracts.ts`, producing a dirty working tree after every build.

The drift was introduced by two prior PRs that hand-edited `contracts.ts` without updating the generator:

- **#424** `feat(webui): bring Server Config tab to parity with macOS tray` — added `Server.isolation_defaults`, `IsolationConfig.network_mode`, `IsolationConfig.extra_args`, and the entire `IsolationDefaults` interface.
- **#463** `feat(webui): per-tool enable/disable + bulk Enable All/Disable All` — added `Tool.disabled` and `Tool.approval_status`.

Neither PR touched `cmd/generate-types/main.go`. The banner at the top of `contracts.ts` says `DO NOT EDIT - This file is auto-generated by cmd/generate-types`, which made it natural to assume re-running the generator would just *produce* those fields. It doesn't — the generator's output is hand-maintained Go string literals, so manual edits to `contracts.ts` need a matching update to the generator or the next build silently reverts them.

### 1. Sync the generator (`cmd/generate-types/main.go`)

Catches the hardcoded TS string literals up to the current `contracts.ts` content. Adds the six missing fields and the `IsolationDefaults` interface so the generator's output is byte-identical to the committed `contracts.ts`.

Also refactors `main()` to extract `generateFileContent()` so the new test can validate the output without re-implementing the header concat.

### 2. Drift-prevention test (`cmd/generate-types/main_test.go`)

`TestContractsInSync` reads the committed `frontend/src/types/contracts.ts`, computes what `cmd/generate-types` *would* write today, and asserts byte-equality. The next time someone hand-edits one side without updating the other, CI fails with a clear message:

> Either run `go run ./cmd/generate-types` from the module root (if the generator is the source of truth) or update the string literals in main.go (if contracts.ts is the source of truth).

This is the smallest discipline that would have caught both #424 and #463 before merge.

## What this PR is *not*

This PR fixes the `contracts.ts` half of the dirty-tree problem only. The committed `web/frontend/dist/*` bundle also churns on every `make build`, but for an independent reason: the bundle in `main` is older than the frontend source it claims to be built from (last refresh was in #433 wizard v2; subsequent feature PRs #424 / #463 added Vue code but didn't refresh the bundle), so Vite produces different content-hashed filenames than the ones committed. That deserves its own discussion (PLACEHOLDER + embed `all:` vs deterministic re-bundle vs leave-as-tracked-artifact) and a separate PR.

## Verification

```
$ go test ./cmd/generate-types/
ok  	github.com/smart-mcp-proxy/mcpproxy-go/cmd/generate-types	0.006s

$ go run ./cmd/generate-types
Successfully generated TypeScript types: frontend/src/types/contracts.ts
$ git diff frontend/src/types/contracts.ts
(empty)

$ go build ./...
(succeeds)
```

Sanity-checked drift detection by appending one line to `contracts.ts` and re-running `go test ./cmd/generate-types/` — test fails with the expected message, then passes again after reverting.

## Test plan
- [x] `go test ./cmd/generate-types/` passes
- [x] `go run ./cmd/generate-types` is idempotent against committed `contracts.ts`
- [x] Test catches a simulated drift (appended line to contracts.ts → test fails as designed)
- [x] `go build ./...` succeeds


Dumbris:
Thank you for this, @electrolobzik! 🙏 This was a real, recurring papercut — every `make build` silently reverting 22 lines and dirtying the tree — and the `TestContractsInSync` guard is exactly the right discipline to stop it recurring.

One snag: the new byte-exact test failed on `windows-amd64` CI because `core.autocrlf=true` checks out `contracts.ts` as CRLF while the generator emits LF. Since this PR is from a fork branch we can't push to, we've carried **your two commits unchanged (authorship preserved)** plus a CRLF fix (`.gitattributes` pin + EOL normalization in the test) into #475, which is now green on all platforms and will ship in v0.31.1. Closing in favor of that — full credit to you for the fix. Much appreciated! 🎉

Dumbris (approved):

PR #520, opened 2026-05-23
fix: filter direct-mode tools by agent token scope

muxammadreza:
## Summary

Direct mode already enforces agent-token scope at tool call time, but tools/list still exposed every direct-mode upstream tool definition. This meant scoped agent tokens blocked execution, but still disclosed tool names, descriptions, and schemas for servers outside the token scope.

This PR filters direct-mode tools/list responses for agent tokens using the same server and permission semantics already used by direct-mode execution.

## Changes

- Track direct-mode tool permission metadata while rebuilding the direct-mode tool registry.
- Add a tools/list filter for auth.AuthTypeAgent contexts.
- Hide tools from servers outside the agent token's AllowedServers scope.
- Hide tools requiring permissions not present on the agent token.
- Preserve existing behavior for admin, OAuth user, and unauthenticated/back-compat contexts.
- Keep call-time authorization unchanged and authoritative.
- Add focused unit tests for server scope, permission scope, fail-closed missing metadata, and non-agent compatibility.

## Verification

- go test ./internal/server -run TestFilterDirectModeToolsForAuth -count=1 -v
- go test ./internal/server -run TestDirectModeHandler_PermissionDenied -count=1 -v
- go test ./internal/auth ./internal/contracts -count=1
- go build ./cmd/mcpproxy
- IntelliJ build_project: passed
- Smoke run on 127.0.0.1:18080 with isolated data dir: /health returned 200 {"status":"ok"}

Note: broader go test ./internal/server and go test ./internal/... attempts exceeded the IDE tool timeout in this local environment before producing Go failures.

### DIFF
--- a/internal/server/mcp.go
+++ b/internal/server/mcp.go
@@ -8,6 +8,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -108,6 +109,13 @@ type MCPProxyServer struct {
 
 	// Hooks shared across all routing mode servers
 	hooks *mcpserver.Hooks
+
+	// directToolPerms maps direct-mode tool names (server__tool) to the
+	// operation permission required to call them. It is populated with the
+	// direct-mode registry and used only to filter tools/list for scoped agent
+	// tokens; execution-time authorization remains authoritative.
+	directToolPermsMu sync.RWMutex
+	directToolPerms   map[string]string
 
 	// Spec 049: in-memory only counter of retrieve_tools calls that opted into
 	// include_disabled. Never persisted (privacy, consistent with Spec 042).
--- a/internal/server/mcp_direct_scope.go
+++ b/internal/server/mcp_direct_scope.go
@@ -1,0 +1,88 @@
+package server
+
+import (
+	"context"
+
+	"github.com/mark3labs/mcp-go/mcp"
+
+	"github.com/smart-mcp-proxy/mcpproxy-go/internal/auth"
+	"github.com/smart-mcp-proxy/mcpproxy-go/internal/config"
+	"github.com/smart-mcp-proxy/mcpproxy-go/internal/contracts"
+)
+
+// requiredPermissionForDirectTool derives the agent-token permission a direct
+// tool requires from its annotations. It reuses the same variant->operation-type
+// mapping that call-time authorization uses (see handleDirectToolCall in
+// mcp_routing.go) so discovery filtering can never diverge from execution
+// enforcement.
+func requiredPermissionForDirectTool(annotations *config.ToolAnnotations) string {
+	return contracts.ToolVariantToOperationType[contracts.DeriveCallWith(annotations)]
+}
+
+func (p *MCPProxyServer) setDirectToolPermissions(perms map[string]string) {
+	p.directToolPermsMu.Lock()
+	defer p.directToolPermsMu.Unlock()
+
+	if len(perms) == 0 {
+		p.directToolPerms = nil
+		return
+	}
+
+	copied := make(map[string]string, len(perms))
+	for name, perm := range perms {
+		copied[name] = perm
+	}
+	p.directToolPerms = copied
+}
+
+func (p *MCPProxyServer) lookupDirectToolPermission(directName string) (string, bool) {
+	p.directToolPermsMu.RLock()
+	defer p.directToolPermsMu.RUnlock()
+
+	perm, ok := p.directToolPerms[directName]
+	return perm, ok
+}
+
+// filterDirectModeToolsForAuth filters tools/list for scoped agent tokens.
+//
+// Direct mode registers upstream tools globally as server__tool. Without this
+// filter, scoped agent tokens prevent execution but still disclose tool names,
+// descriptions, and schemas for servers outside their scope. Call-time auth is
+// still authoritative; this filter only removes tools that the current token
+// could not call from discovery responses.
+func (p *MCPProxyServer) filterDirectModeToolsForAuth(ctx context.Context, tools []mcp.Tool) []mcp.Tool {
+	if len(tools) == 0 {
+		return tools
+	}
+
+	authCtx := auth.AuthContextFromContext(ctx)
+	if authCtx == nil || authCtx.Type != auth.AuthTypeAgent {
+		return tools
+	}
+
+	filtered := make([]mcp.Tool, 0, len(tools))
+	for _, tool := range tools {
+		serverName, _, ok := ParseDirectToolName(tool.Name)
+		if !ok {
+			filtered = append(filtered, tool)
+			continue
+		}
+
+		if !authCtx.CanAccessServer(serverName) {
+			continue
+		}
+
+		requiredPerm, ok := p.lookupDirectToolPermission(tool.Name)
+		if !ok {
+			continue
+		}
+
+		if requiredPerm != "" && !authCtx.HasPermission(requiredPerm) {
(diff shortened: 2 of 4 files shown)


muxammadreza:
The actual review follow-up in commit b769a55 is:

- `requiredPermissionForDirectTool` now returns `auth.Perm*` constants explicitly instead of relying on string equality with `contracts.OperationType*`.
- `setDirectToolPermissions` defensively copies the permission map before publishing it.
- `filterDirectModeToolsForAuth` now allocates a fresh filtered slice instead of reusing the caller's backing array.
- Added focused tests for annotation-derived permission mapping, defensive map copy behavior, and non-mutating filter behavior.

Verification:

- `go test ./internal/server -run TestRequiredPermissionForDirectTool -count=1 -v`
- `go test ./internal/server -run TestSetDirectToolPermissions -count=1 -v`
- `go test ./internal/server -run TestFilterDirectModeToolsForAuth_DoesNotMutateInputSlice -count=1 -v`
- `go test ./internal/server -run TestFilterDirectModeToolsForAuth -count=1 -v`
- `go build ./cmd/mcpproxy`
- IntelliJ `build_project` passed

Dumbris (commented):
Thanks for this — the agent-token scoping half is genuinely solid work. The `directToolPerms` map under an `RWMutex`, the defensive copy, the fail-closed-on-missing-metadata path, and the thorough table tests (server scope / permission scope / non-mutation / admin passthrough) are exactly right. Closing the `tools/list` info-disclosure leak for scoped tokens is a real security win.

I'd like to talk through one thing before we land it, because it touches our core security model.

## The PR actually contains two independent changes

1. **Agent-token scope filtering** — the stated purpose. Read-time only, no persistence, no migration. 👍
2. **Output schema folded into the tool hash** (`internal/hash/hash.go`, `ToolMetadata.OutputSchemaJSON`, index, `upstream/core`) — not mentioned in the description, and it changes on-disk behavior.

These are orthogonal (the filter never touches a hash), so I'd ask that they be split into two PRs regardless of what we decide below — the migration deserves its own title and review so a future bisect of "why did all my tools re-quarantine after upgrade" lands on the right commit.

## The contract model this needs to respect

Our tool-quarantine (Spec 032) encodes a specific policy:

- A **human** approves a server's tool contract — description + schema — **once**.
- Agents may use **only approved tools**. Agent tokens cannot self-approve; they're blocked at call time until a human acts.
- If the **upstream server** later changes a tool's description or schema, that tool is **re-quarantined** (rug-pull guard) until a human re-approves.

The change-detection hash is the fingerprint of "the contract the human signed off on."

## Why this matters for the output-schema change

Two parts, and they pull in opposite directions:

**(a) Output schema *should* be in the hash.** Output schema is part of the contract — it describes what the agent receives back. A server silently swapping its output shape is a contract change a human should re-review. So I agree with the intent: keep output schema in the hash. Dropping it would let output-schema rug-pulls slip past review.

**(b) But the upgrade itself must not re-quarantine anything.** Here's the subtlety: when a user upgrades to this build, the *upstream contract is byte-identical* — only **mcpproxy's hashing algorithm** changed. As written, every already-approved tool that ships an output schema will hash differently and flip to `changed`, demanding human re-approval for tools that **did not actually change upstream.**

That's a false positive, and it's the dangerous kind: the first time a user sees "all 200 tools quarantined, click Approve-All," they stop reading the diffs. That trains humans to rubber-stamp — which destroys the exact signal the whole policy depends on.

## Proposed solution

1. **Keep output schema in the hash** — it's part of the approvable contract.
2. **Add a one-time versioned backfill** so the algorithm change never masquerades as a rug-pull. We already have the hook: `CurrentSchemaVersion` / `SchemaVersionKey` in `internal/storage/models.go`. On boot, when stored version < N:
   - walk every **`approved`** `ToolApprovalRecord` (and the index / `ToolHashBucket` hashes),
   - recompute the hash with the new output-schema-inclusive algorithm using the **currently observed** output schema,
   - rewrite `ApprovedHash`/`CurrentHash` while **preserving `Status = approved`**.
   - leave `pending` / `changed` records untouched — they still need human action.

   Semantics: *"the contract this human already approved hasn't changed; we're only re-baselining the fingerprint to include output schema."* From that point on, any real drift in description, input schema, **or** output schema → `changed` → human-only re-approval. That's the policy, with no exception carved out.

   One honest caveat we should document: tools approved before the upgrade have no prior output-schema baseline, so we can't retroactively detect drift that happened *before* the migration — the backfill blesses whatever output schema exists at migration time. Unavoidable (no historical baseline exists), and acceptable since the server was already trusted at the contract the human reviewed. The window exists only once, at the upgrade boundary.

3. **(Follow-on, optional but it falls right out of this PR)** "Agents may use only approved tools" implies the agent `tools/list` filter you added should *also* hide `pending`/`changed` tools, not just out-of-scope ones — discovery should match capability, so an agent sees exactly the set it can invoke (in-scope ∧ permitted ∧ approved). Human/admin/OAuth-user contexts still see pending/changed, since they're the ones who act on them.

## Suggested shape

- **PR A** (this one, trimmed): agent-token scope filter only. Mergeable as soon as the full `go test ./internal/server/...` run is confirmed green (the description notes only the targeted `-run` filters were run locally).
- **PR B**: output schema in hash + the versioned backfill migration + a test proving an approved output-schema tool stays approved across a simulated version bump.
- **PR C** (optional): extend the agent filter with the approval-status predicate.

If you're good with this direction, I'd be happy to pair on it — I can take the migration in PR B (or push directly to your branch) while you keep ownership of the scoping work. Let me know what you'd prefer. Thanks again — the hard part is already done well.

muxammadreza:
Agreed on your points, and regarding to PR C, I'm gonna split it (I originally noticed the issue and wanted to combine it in this PR, but as you suggested, I'll keep it for another fresh PR).

## Summary

Filters direct-mode `tools/list` responses for scoped agent tokens.

Direct mode already enforced agent-token server and permission scope at tool-call time, but discovery still exposed every registered direct-mode upstream tool. This change applies the same server/permission model during `tools/list` so scoped agents only see tools they are allowed to call.

## Changes

- Adds direct-mode tool permission metadata keyed by `server__tool`.
- Installs an `mcp-go` `WithToolFilter` on the direct-mode MCP server only.
- Filters only `auth.AuthTypeAgent`; admin, user, and unauthenticated contexts preserve existing behavior.
- Fails closed if direct tool permission metadata is missing.
- Keeps execution-time authorization unchanged and authoritative.
- Adds tests for permission derivation, server scope, permission scope, non-mutation, missing metadata, and admin/non-agent passthrough.

## Out of scope

This PR intentionally does not include output-schema hashing or storage/index migration work. That behavior is orthogonal and should be reviewed separately.

Approval/quarantine/disabled-state filtering for agent discovery and call-time enforcement is also left for a follow-up PR.

## Validation

- `go test ./internal/server -run TestFilterDirectModeToolsForAuth -count=1`
- `go test ./internal/server -run 'TestHandleCallToolVariant|TestHandleUpstreamServers|TestHandleListUpstreams|TestHandleQuarantineSecurity|TestFilterDirectModeToolsForAuth' -count=1`
- `go test ./internal/server -run TestDoesNotExist`
- pre-commit hooks: passed
- pre-push hooks: passed, including OpenAPI verification and `go build`

Dumbris (approved):
Rebased onto current main (picks up #526/#529). Applied review fixes: both direct-mode agent filters now live on directOpts only; requiredPermissionForDirectTool reuses contracts.ToolVariantToOperationType so discovery filtering can't drift from call-time enforcement. Full internal/server -race suite green locally; all CI checks pass.

Issue #333, opened 2026-03-10
feat: expand secret/env refs in working_dir (and data_dir, host_path)


tjsingleton:
## Summary

`${env:VAR}` and `${keyring:...}` references are expanded for `env`, `args`, and `headers` in `NewClientWithOptions` ([`internal/upstream/core/client.go`](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/f204be8c42648b20b7b1eeee8d55836f239c2c9f/internal/upstream/core/client.go)), but `working_dir` is never passed through `ExpandSecretRefs`. The same gap exists for the top-level `data_dir` config field and Docker isolation `host_path` volume mounts.

## Reproduction

Config:
```json
{
  "working_dir": "${env:IH_HOME}",
  ...
}
```

The literal string `${env:IH_HOME}` is passed as the working directory instead of the resolved path, causing the server to fail to start if the directory doesn't literally exist.

## Expected Behavior

`${env:...}` and `${keyring:...}` references in `working_dir`, `data_dir`, and `host_path` are resolved the same way they are for `env`, `args`, and `headers`.

## Suggested Fix

In `NewClientWithOptions`, add a resolution block for `WorkingDir` after the existing `Headers` block:

```go
// Resolve secrets in working directory
if resolvedServerConfig.WorkingDir != "" {
    resolvedValue, err := secretResolver.ExpandSecretRefs(ctx, resolvedServerConfig.WorkingDir)
    if err != nil {
        logger.Error("Failed to resolve secret in working_dir", ...)
    } else {
        resolvedServerConfig.WorkingDir = resolvedValue
    }
}
```

Similar fixes needed for:
- Top-level `Config.DataDir`
- `IsolationConfig.Volumes[].HostPath`

## Context

Validated against commit `f204be8`.

Issue #56, opened 2025-09-13
Created a lot of processes on MacOS


tzutoo:
I have only 3 MCPs setup. However, it creates a lot of processes for each of the MCP.

tzutoo:
2025-09-13T15:41:48.522+08:00 | INFO | managed/client.go:88 | Starting connection attempt | {"server": "zai-mcp-server", "transport": "stdio", "url": "", "command": "npx", "protocol": ""}
2025-09-13T15:42:18.522+08:00 | ERROR | core/connection.go:276 | MCP initialize JSON-RPC call failed | {"server": "zai-mcp-server", "error": "transport error: context deadline exceeded"}
2025-09-13T15:42:18.523+08:00 | ERROR | managed/client.go:88 | Connection failed | {"server": "zai-mcp-server", "transport": "stdio", "error": "MCP initialize failed for stdio transport: MCP initialize failed: transport error: context deadline exceeded"}
2025-09-13T15:42:18.524+08:00 | INFO | managed/client.go:88 | Starting connection attempt | {"server": "zai-mcp-server", "transport": "stdio", "url": "", "command": "npx", "protocol": ""}
2025-09-13T15:42:28.525+08:00 | ERROR | core/connection.go:276 | MCP initialize JSON-RPC call failed | {"server": "zai-mcp-server", "error": "transport error: context deadline exceeded"}
2025-09-13T15:42:28.525+08:00 | ERROR | managed/client.go:88 | Connection failed | {"server": "zai-mcp-server", "transport": "stdio", "error": "MCP initialize failed for stdio transport: MCP initialize failed: transport error: context deadline exceeded"}
2025-09-13T15:42:35.660+08:00 | INFO | managed/client.go:88 | Starting connection attempt | {"server": "zai-mcp-server", "transport": "stdio", "url": "", "command": "npx", "protocol": ""}
2025-09-13T15:43:05.661+08:00 | ERROR | core/connection.go:276 | MCP initialize JSON-RPC call failed | {"server": "zai-mcp-server", "error": "transport error: context deadline exceeded"}
2025-09-13T15:43:05.661+08:00 | ERROR | managed/client.go:88 | Connection failed | {"server": "zai-mcp-server", "transport": "stdio", "error": "MCP initialize failed for stdio transport: MCP initialize failed: transport error: context deadline exceeded"}


---


every failure creates a new process

Dumbris:
@tzutoo Thank you for opening this issue! 

[...]

Issue #55, opened 2025-09-12
Not all mcp servers are relevant for all clients


technicalpickles:
There are tons of MCP out there that assume they are being run from the directory the MCP client is. 

For example, [ast-grep-mcp](https://github.com/ast-grep/ast-grep-mcp) will run in the directory it’s started from, or take an explicit directory to run from.

If a user is running a single mcpproxy (which is encouraged by having things like a menu bar to manage), it could get confusing to add ast-grep-mcp to use in one project, but then another project ends up trying to call it and gets results for the wront project.

Another example is having a private mcp server for your work, that simply doesn’t apply to projects not in your company’s github org.

I have  a few ideas, but wanted to try to describe the general problem I’d like to solve.

Dumbris:
@technicalpickles Thanks for raising this important issue!

  My understanding of the issue (please correct me if I'm wrong): When running mcpproxy as a global daemon, MCP servers that operate on files/directories (like ast-grep-mcp, filesystem, git servers) run from mcpproxy's working directory instead of your actual project directories. This causes servers to search/operate in the wrong location and creates issues when working across multiple projects.

  Potential solution: We'll add a working_dir field to the server configuration, allowing you to specify exactly where each MCP server should run from:

```
  {
    "mcpServers": [
      {
        "name": "ast-grep-project-a",
        "command": "uv",
        "args": ["run", "ast-grep-mcp"],
        "working_dir": "/home/user/projects/project-a",
        "enabled": true
      },
      {
        "name": "ast-grep-work",
        "command": "uv",
        "args": ["run", "ast-grep-mcp"],
        "working_dir": "/home/user/work/company-repo",
        "enabled": true
      }
    ]
  }
```

And this is  already available for Docker isolation: If you're using [Docker isolation](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/main/docs/docker-isolation.md), working directory support is already partially implemented:

```

  {
    "docker_isolation": {
      "enabled": true,
      "memory_limit": "512m",
      "cpu_limit": "1.0"
    },
    "mcpServers": [
      {
        "name": "ast-grep-secure",
        "command": "npx",
        "args": ["ast-grep-mcp"],
        "isolation": {
          "enabled": true,
          "working_dir": "/home/user/projects/secure-project"
        },
        "enabled": true
      }
    ]
  }
```

[...]

PR #531, opened 2026-05-26
docs(spec): add spec 057 for in-proxy profiles (#55)

Ylsssq926:
Captures the design discussion from #55 (@Dumbris's "In-Proxy Profiles + Permanent URLs" plan plus the scope trade-offs from the follow-up comments) into a draft spec, so the implementation can be reviewed in two stages, matching the pattern from #525 / #056.

## What this is

Just the spec doc: \`specs/057-in-proxy-profiles/spec.md\` (single-file draft, no \`plan.md\` / \`tasks.md\` yet; those come after direction is locked).

## Scope captured in the spec

**MVP (this spec, next PR):**
- \`config.profiles: [{name, servers}]\` with slug-validated names
- New permanent URL \`/mcp/p/{slug}\`, stateless, filters the visible upstream set per request
- Filter wired into the existing scope hooks in \`handleRetrieveTools*\` / \`handleCallToolVariant\`, composes with agent-token \`AllowedServers\` (Spec 028) by intersection
- Reuses per-server \`enabled_tools\` / \`disabled_tools\` for tool-level filtering inside a profile
- Zero migration: missing \`profiles\` keeps current behaviour

**Out of scope (deferred, with reasons in the spec):**
- Active profile switching (state ownership is an open question; \`/mcp\` semantics stay as-is)
- Tray UI / \`set_profile\` MCP tool (depend on active state)
- Index-level profile field (cardinality is small, filter at search-result time)
- Melodeiro's \`server:tool\` mixed list (existing \`enabled_tools\` already expresses it)

## Open Questions for review

Three calls I deliberately left for you to make in the spec (rather than baking them in):

1. \`profile.servers\` referencing an unknown server: hard error or warn-and-skip?
2. \`/mcp\` semantics when \`profiles\` is configured: keep "show everything" (current behaviour) or expose a "default" profile?
3. Should \`all\` be reserved as a slug for a future "show everything" profile?

Refs: #55, #333

### DIFF
--- a/specs/057-in-proxy-profiles/spec.md
+++ b/specs/057-in-proxy-profiles/spec.md
@@ -1,0 +1,265 @@
+# Feature Specification: In-Proxy Profiles + Permanent URLs
+
+**Feature Branch**: `057-in-proxy-profiles`
+**Created**: 2026-05-26
+**Status**: Draft. Awaiting maintainer review on direction. Implementation tracked separately.
+**Input**: Issue #55. Reporter @technicalpickles asked for two related capabilities: per-server `working_dir` (already shipped via `ServerConfig.WorkingDir`, related issue #333), and a way for different MCP clients to see different subsets of upstream servers from the same proxy instance. @Dumbris responded with the design "In-Proxy Profiles + Permanent URLs". @Melodeiro suggested an extension that mixes `server` and `server:tool` entries in profile lists.
+
+> Scope note: this spec covers the **MVP** of profiles only. The MVP is a stateless URL-based selector. Active-profile switching, a tray selector, a `set_profile` MCP tool, and an indexable `profile` field are explicitly deferred (see [Out of Scope](#out-of-scope)). The deferred items depend on resolving an open question about where active-profile state lives (per-process, per-session, or per-token), which is parked under [Open Questions](#open-questions) for the maintainer to direct.
+
+## User Scenarios & Testing *(mandatory)*
+
+### User Story 1 - Two clients, two profiles, one proxy (Priority: P1)
+
+An operator uses a single MCPProxy instance to back several MCP clients on the same machine. Some clients work on a "research" set of servers (web search, scratch filesystem), others on a "deploy" set (GitHub, Kubernetes, internal CI). They want each client connected to a permanent URL that exposes only its profile's servers, rather than seeing the full union of every configured server.
+
+**Why this priority**: This is the core ask in #55. Without it, every client sees every server, which leaks unrelated tools into each agent's context and makes BM25 retrieval less precise on a per-client basis. It is a pure addition with no impact on existing setups.
+
+**Independent Test**: Configure two profiles (`research` and `deploy`) referencing distinct subsets of servers. Connect Client A to `/mcp/p/research` and Client B to `/mcp/p/deploy`. Confirm `retrieve_tools` on Client A only returns tools from `research`'s servers, and `call_tool_*` on Client A rejects calls into a `deploy`-only server. Confirm `/mcp` (no profile) still returns the full union.
+
+**Acceptance Scenarios**:
+
+1. **Given** a config with `profiles: [{name: "research", servers: ["web", "fs"]}, {name: "deploy", servers: ["github", "k8s"]}]`, **When** a client connects to `/mcp/p/research` and calls `retrieve_tools`, **Then** only tools from `web` and `fs` are returned.
+2. **Given** the same config, **When** a client connects to `/mcp/p/research` and invokes `call_tool_read` for a tool on `github`, **Then** the call is rejected with a clear "server not in profile" error.
+3. **Given** the same config, **When** a client connects to `/mcp` (no profile), **Then** `retrieve_tools` returns tools from all four servers (no behavioural change versus today).
+4. **Given** a config with no `profiles` field at all, **When** a client connects to `/mcp` or to any `/mcp/p/<anything>` URL, **Then** `/mcp` behaves exactly as today and `/mcp/p/<slug>` returns 404 with a body indicating "no profiles configured".
+
+---
+
+### User Story 2 - Profile composes with agent token scope (Priority: P1)
+
+An operator already issues scoped agent tokens (Spec 028) that limit which servers an agent may use. They want to layer profiles on top so a single agent token, used against different profile URLs, naturally narrows further by the URL it is presented at, without having to re-issue tokens. The agent's effective server set is the **intersection** of `AgentToken.AllowedServers` and the profile's `servers`.
+
+**Why this priority**: Profiles and agent tokens are two independent scoping primitives the project already exposes. They MUST compose by intersection to remain predictable. Without this, operators have to choose between the two.
+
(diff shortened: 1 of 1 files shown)


Dumbris (commented):
Thanks for capturing the #55 discussion into a reviewable spec — the staging, the explicit "Out of Scope" reasons, and the agent-token/per-user composition framing are all on point. I did a deep pass against the current codebase to check the design holds before it goes to `plan.md`. The core is compatible; below are the amendments I'd like folded in before we lock direction.

## A. Resolve the three Open Questions (decided)

The "Open Questions" section can become decisions:

1. **Unknown server reference → warn and skip.** Promote FR-015 from "recommendation" to settled; drop the hard-error alternative. Rationale: parity with Spec 028, and config must stay loadable after a server rename.
2. **`/mcp` semantics → always full union, purely additive.** State as a settled invariant, not a fork. No breaking change for existing `/mcp` clients; profiles are opt-in narrowing via the new URL only. (A `strict` opt-in can be a later spec if anyone asks.)
3. **Reserve `all` → yes**, but for the corrected reason in (B).

## B. Fix a factual error about `/mcp/all`

The spec describes `all` as "reserved for a **possible future** explicit-all profile … without committing to its semantics now." That's not accurate — `/mcp/all` is **already a live, bound endpoint**: it serves direct routing mode (Spec 031, `internal/server/server.go:1670`). Suggested correction for Edge Cases + the Spec 031 reference:

> `all`, `code`, and `call` are already bound routing-mode subpaths under `/mcp/` (Spec 031); `p` is the profile prefix itself. These slugs are reserved to avoid operator confusion — note there is no actual path collision, since profiles live under the distinct `/mcp/p/` prefix.

## C. Add an "Implementation Design" section (current biggest gap)

The spec says profiles are "wired into the existing scope hooks" but never says *how*. Two findings need to be in the spec because an implementer would otherwise get them wrong:

1. **Routing mechanism — middleware + context, not per-profile server instances.** Routing modes (`/mcp/all|code|call`) register a *separate MCP server instance per path at startup* via `GetMCPServerForMode`. That can't be mirrored for N hot-reloadable profile slugs because `http.ServeMux` can't de-register routes. The framework-friendly fit is a single `/mcp/p/` prefix handler that strips the slug, looks up the profile, and injects the resolved server set into the request context — exactly how `mcpAuthMiddleware` injects `AuthContext` — reusing the existing MCP server instance. Middleware order: auth → profile.

2. **The filter must run independently of auth type (correctness-critical).** The existing server-scope filter in `retrieve_tools` (`mcp.go` ~1108) and `call_tool_*` (`mcp.go` ~1491) is gated on `enforceAgentScope := authCtx != nil && !authCtx.IsAdmin()`. But a default `/mcp/p/...` connection has **no token and is assigned `AdminContext()`**, for which that gate is `false` and `CanAccessServer` returns `true` unconditionally. So profile filtering **cannot** ride the agent-scope gate or be implemented by stuffing the profile's servers into `AuthContext.AllowedServers`. It must be a *parallel* check that runs for every auth type:

   ```go
   if profileScope != nil && !profileScope.Allows(serverName) { /* hide / reject-with-profile-error */ }
   if enforceAgentScope && !authCtx.CanAccessServer(serverName) { /* existing token gate */ }
   ```

   Two independent checks give the intersection (FR-005) and the two distinct error messages (FR-012) for free, with **no change to `AuthContext`, the `agent_tokens` bucket, or token validation**. Please add an explicit FR for this and a named regression test: *"an unauthenticated connection at `/mcp/p/<slug>` is still filtered to the profile's servers."*

3. **Config store + reload.** Top-level `profiles` array in the config file, hot-reloaded. `Profiles []ProfileConfig \`json:"profiles,omitempty"\`` → byte-identical round-trip when absent (SC-004 for free).

4. **Files touched (scope guard).** `internal/config` (+ new `profiles.go` validation), new `internal/profile/context.go` (~30 lines, mirrors `auth/context.go`), `internal/server/server.go` (one route + `profileMiddleware`), `internal/server/mcp.go` (two filter conditions + one metadata write). Explicitly **no** storage, index, or token-model changes. Per-server `enabled_tools`/`disabled_tools` need zero changes — they already apply downstream of the server gate, so FR-006/US3 works automatically.

## D. Tighten FR-011

Specify that the profile slug lands in the existing `ActivityRecord.Metadata` map (`metadata["profile"]=slug`), not a new top-level field — no schema change needed, matching how Specs 018/026 attach context.

---

Net: design is sound and lands on the right seams. The must-fix items are **B** (factual) and **C.2** (latent correctness bug). A, C.1/3/4, and D are clarifications that make the spec safe to implement directly. Happy to pair on the wording if useful.

Dumbris:
**Critic (Codex) review — Ylsssq926's PR #531**
Verdict: accept
Strengths: The PR is spec-only (`specs/057-in-proxy-profiles/spec.md`), captures the profile URL design and explicit out-of-scope decisions, and all checks are green at head `cf9218162df319675938872275197e3f9c43aef4`.
Findings: none.
Provenance check: ok

Dumbris:
**Critic (Codex) review — Ylsssq926's PR #531**
Verdict: accept
Head: `cf9218162df319675938872275197e3f9c43aef4`

Strengths: Spec-only change with a clear MVP boundary, explicit composition rules for profiles with agent tokens/user visibility, and documented open questions rather than hidden assumptions. CI is green for all non-skipped checks.

Findings: none

Provenance check: ok

Dumbris:
## Review — accept & merge (spec-only)

Verified the load-bearing technical claims against the codebase and they hold:
- `enforceAgentScope := authCtx != nil && !authCtx.IsAdmin()` (`internal/server/mcp.go:1113`), gate at `:1142` / call-path at `:1529`.
- Routing modes register a separate server instance per static `mux.Handle` (`server.go:1670-1694`), so the spec's claim that this can't be mirrored for hot-reloadable profiles (ServeMux can't de-register routes) is correct — middleware+context is the right seam.
- The key insight is right and important: an unauthenticated `/mcp/p/` connection gets `AdminContext()`, where `enforceAgentScope` is false and `CanAccessServer` returns true unconditionally. So profile filtering **must** be a parallel check, not ride the agent-scope gate. The mandatory named regression test for this is the right call.

On-vision: directly answers #55, composes by intersection with agent tokens (028) and per-user visibility, zero migration, `/mcp` untouched. Good scope discipline deferring active-profile state.

Two product reconciliations to settle **before the implementation PR** (not blockers for this spec doc):
1. **URL scheme** — spec ships `/mcp/p/<slug>`; roadmap/blog promise `/mcp/<profile>`. The spec's choice is the correct engineering call (avoids the `/mcp/all|code|call` collision); marketing copy should be updated to match.
2. **No UI in the MVP** — profiles are config-file-only; the launch narrative features a tray switcher that's explicitly out-of-scope. Recommend a Web UI "Profiles" page as the immediate fast-follow (CRUD over `profiles[]`, copyable `/mcp/p/<slug>` URL) — it needs no active-state decision and is what sells "zero config."

Minor: confirm the per-user-visibility reference number (spec says 029; multiuser work is tracked as 024 elsewhere). Merging the spec.

PR #560, opened 2026-06-01
feat(069-A2): actor-owned usage aggregate + persistence

Dumbris:
## Spec 069 Stream A2 (Backend) — usage aggregate + persistence

Implements tasks **T005–T010** from `specs/069-observability-usage-graphs/tasks.md`. Builds on A1 (#553, byte capture). Unblocks A3 (`/api/v1/activity/usage` endpoint, MCP-750).

### What landed
- **`internal/runtime/usage_aggregate.go`** — `UsageAggregate` / `ToolUsage` / `TimeBucket` with incremental `Apply(record)`. `UsageStore` publishes an immutable deep copy via `atomic.Pointer` (copy-on-write); `Snapshot()` reads are lock-free and never block.
- **`ActivityService` ownership** — the aggregate is mutated only on the ActivityService goroutine (`Apply` on save success in `handleEvent`), exposed via `UsageSnapshot()`. Periodic flush (default 30s) + flush-on-shutdown; cold start does **exactly one** full-scan rebuild when no snapshot exists.
- **`internal/storage/activity_stats.go`** — new `activity_stats` BBolt bucket, versioned key `usage_aggregate_v1`. Byte-oriented `SaveUsageSnapshot`/`LoadUsageSnapshot` + `ScanAllActivities` (avoids a storage→runtime import cycle; runtime owns JSON encode/decode + orchestration).
- **Config** — `observability.usage_cache_ttl` (5s) + `usage_persist_interval` (30s): defaults in `DefaultConfig`, repaired in `Validate` (hot-reload re-runs it), and the flush loop re-reads the interval each cycle.

### Design decisions (informed assumptions — see `data-model.md` §2 "A2 implementation notes")
- **Hourly time buckets** (matches the contract's `start: …T11:00:00Z`), bounded ring of `24*90` buckets (~90d retention), oldest evicted. Minute/5-min 24h granularity deferred — hourly bars render cleanly and keep the aggregate bounded. A3 trims/selects the window over these.
- **Separate `SizedRespCalls`/`SizedReqCalls`** so `avg_resp_bytes`/`avg_req_bytes` each exclude their own 0-byte (legacy) records exactly. Contract `sized_calls` → `SizedRespCalls`.
- **Fixed latency histogram** (`{10,25,50,100,250,500,1000,2500,5000,10000}` + overflow); `Percentile(p)` returns the bucket upper bound.

### Verification (ENG-7/ENG-8)
- TDD: failing tests first for aggregate math, snapshot/reads-never-block (incl. concurrent readers under `-race`), persistence round-trip, cold-start load-vs-rebuild, handleEvent integration, and config defaults.
- `go test ./internal/runtime/... ./internal/storage/... ./internal/config/... -race` ✅ (full runtime suite incl. `TestCalculateToolApprovalHash_Stability` canary — `ActivityRecord` consumers unchanged).
- `./scripts/run-linter.sh` → 0 issues. `gofmt` clean. Personal + `-tags server` builds ✅.
- Docs: `docs/configuration.md` Observability section; `oas/swagger.yaml` regenerated (config schema). No REST endpoint added (that's A3).

Related #745


Dumbris:
**Critic (Codex) review — Dumbris's PR #560**
Verdict: request_changes
Strengths: The PR wires config, docs, persistence, and tests for the actor-owned usage aggregate, and all GitHub checks are green at head `62523fe2f82d4d69c68bcec21a19c3311e7fbaaf`.
Findings:
- `internal/runtime/usage_aggregate.go:275` / `internal/runtime/usage_aggregate.go:292`: `UsageStore.Apply` republishes on every activity record, and `publishLocked` deep-clones the full aggregate (`working.clone()`) each time. That makes the ActivityService write path O(number of tools + buckets) per saved record, not the O(1) hot-path update promised by Spec 069 CN-002/FR-005 and by the PR description. With high-cardinality tools or many retained buckets, a single tool-call activity write can copy the whole 90-day aggregate while holding the store mutex. The snapshot publication needs to be decoupled/batched or otherwise bounded so activity writes remain incremental.
Provenance check: ok

Dumbris:
**Critic (Codex) review — Dumbris's PR #560**
Verdict: request_changes
Head: `62523fe2f82d4d69c68bcec21a19c3311e7fbaaf`

Strengths: The aggregate is scoped to the A2 backend stream, uses copy-on-write snapshots for readers, adds persistence tests, and updates config docs/OpenAPI for the new observability fields.

Findings:

1. Blocked tool calls will be missing from the usage aggregate. The usage endpoint contract exposes `status=blocked` filtering and a per-tool `blocked` count (`specs/069-observability-usage-graphs/contracts/usage-endpoint.md:14`, `specs/069-observability-usage-graphs/contracts/usage-endpoint.md:37`). In the current runtime, blocked tool attempts are emitted via `EmitActivityPolicyDecision` (`internal/runtime/event_bus.go:451`) and the server call sites use that path for quarantine/validation blocks (`internal/server/mcp.go:1571`, `internal/server/mcp.go:1613`, `internal/server/output_validation.go:13`). `ActivityService` persists those as `ActivityTypePolicyDecision` records (`internal/runtime/activity_service.go:204`, `internal/runtime/activity_service.go:347`), but `UsageAggregate.Apply` ignores anything that is not `ActivityTypeToolCall` (`internal/runtime/usage_aggregate.go:153`). The test suite even locks in that policy decisions with `Status: "blocked"` are ignored (`internal/runtime/usage_aggregate_test.go:58`). Result: the A3 endpoint will report zero blocked calls for the real blocked-call path, breaking the contract/status filter.

2. `observability.usage_persist_interval` is not actually hot-reloaded. The data model says the observability config is hot-reloadable (`specs/069-observability-usage-graphs/data-model.md:67`), and `SetUsagePersistInterval` is documented as hot-reloadable (`internal/runtime/usage_service.go:55`). But the runtime only calls it during construction (`internal/runtime/runtime.go:209`), while config hot-reload change detection never includes `observability` (`internal/runtime/config_hotreload.go:83`) and `ApplyConfig` only applies logging/tool-response-limit component changes before updating the config snapshot (`internal/runtime/runtime.go:1250`, `internal/runtime/runtime.go:1262`). Editing `usage_persist_interval` at runtime therefore saves the config but leaves the existing flush loop on its old interval.

Checks: `gh pr checks 560 --watch=false` showed no failing or pending checks.

Provenance check: ok

Dumbris:
Addressed CodexReviewer's hot-path regression finding in commit `65540ab1`: `UsageStore.Apply` previously deep-cloned the whole aggregate on every activity write (publish-on-write → O(tools×buckets)). Now `Apply` mutates under a short writer lock + atomic `dirty` flag only (O(1), no per-write clone); the clone is deferred to **publish-on-read** in `Snapshot()` (lock-free fast path when clean). New `TestUsageStore_ApplyDoesNotPublishPerWrite` (500 writes → 0 publishes, 1 clone on first read) + `BenchmarkUsageStore_Apply` (1 alloc/op with 1000 primed tools). Full `internal/runtime -race` suite, build, and linter green locally.

Dumbris:
Commit `17d85375` addresses CodexReviewer's remaining findings (the O(1) fix was #1, cleared once CI went green):

- **#2 blocked counts**: blocked tool attempts are persisted as blocked `policy_decision` records and were never folded into the aggregate (`ToolUsage.Blocked` was permanently 0). `Apply` now folds them — incrementing **only** `Blocked`+`LastUsed`, not `Calls`/latency/bytes/timeline — and `handlePolicyDecision` feeds the aggregate so the live path matches a cold-start rebuild. Contract documents the semantics.
- **#3 observability hot-reload**: `usage_persist_interval` claimed hot-reload but was construction-only. `DetectConfigChanges` now flags `observability` and `ApplyConfig` pushes the new cadence to the running ActivityService.

Tests: `TestUsageAggregate_Apply_CountsBlockedPolicyDecisions`, `TestActivityService_HandleEvent_CountsBlockedPolicyDecision`, `TestDetectConfigChanges_Observability`, `TestApplyConfig_ObservabilityHotReload`. Full `internal/runtime+config+storage -race` green; lint 0; personal+server builds.

PR #581, opened 2026-06-03
ci: de-conflict stable/prerelease workflows for RC builds (MCP-1012)

Dumbris:
## Problem

`release.yml` triggers on tags `v*`, which **also matches RC tags** like `v0.37.0-rc.1`. So an RC tag currently fires BOTH the stable `release.yml` and `prerelease.yml` workflows. The stable workflow's `update-homebrew` and `publish-linux-repos` jobs already guarded with `!contains(github.ref_name, '-')`, but the rest did not — so a stable GitHub Release would be cut for an RC tag and `releases/latest` would point at the RC.

## Fix: de-conflict (release.yml)

Added `&& !contains(github.ref_name, '-')` to the `if:` of the stable-only jobs that lacked it, using the exact idiom already in `update-homebrew`/`publish-linux-repos`:

| Job | Change |
|-----|--------|
| `generate-notes` | guard added |
| `build` | guard added |
| `sign-windows` | guard added |
| `release` | had **no** `if:` — added `if: startsWith(github.ref,'refs/tags/v') && !contains(github.ref_name,'-')` |
| `deploy-docs` | guard added (preserving the `github.repository == ...` condition) |
| `trigger-marketing-update` | guard added (preserving the `github.repository == ...` condition) |

`provenance` (`needs: [release]`) and `mcp-registry` (`needs: release`) cascade-skip when `release` skips. `build-docker` is `if: false`. Net effect: the stable workflow no-ops on any prerelease (`-`) tag; only `prerelease.yml` runs.

## Prerelease completeness (prerelease.yml)

Brought the prerelease build matrix to parity with stable so an RC is a complete downloadable build:

- **Windows now builds natively** on `windows-latest` with `cgo: "1"` (was cross-compiled on `ubuntu-latest` with `cgo: "0"`, where the `choco install innosetup` / `pwsh` installer steps cannot run).
- Builds the **Windows tray binary** (`mcpproxy-tray.exe`) and includes the tray binary in archives (matching stable).
- Uses **PowerShell `Compress-Archive`** on Windows (the `zip` CLI isn't available on Windows runners).
- Added `shell: bash` to the build + frontend-copy steps so they work on the native Windows runner; aligned the frontend embed-copy with stable (`web/frontend/dist` + `.gitkeep`).

Already at parity (unchanged): Linux `.deb`/`.rpm`, macOS DMG/PKG signing + notarization, all 6 platform archives, `prerelease: true`. SignPath Windows-installer signing remains **stable-only** by design — RCs ship an unsigned installer (the release body already states "use at your own risk").

## Validation

- `actionlint` + `python yaml.safe_load` on both files: no syntax/expression errors introduced (remaining reports are pre-existing shellcheck info/style + the `matrix.edition` fallback warnings).
- Tag-push smoke test with `v0.0.0-rc.1` (see PR checks): only `prerelease.yml` should run; `release.yml` jobs skip; `releases/latest` stays on the prior stable; no Homebrew/Linux-repo publish.

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -33,16 +33,16 @@ jobs:
             cgo: "0"
             name: mcpproxy-linux-arm64
             archive_format: tar.gz
-          - os: ubuntu-latest
+          - os: windows-latest
             goos: windows
             goarch: amd64
-            cgo: "0"
+            cgo: "1"
             name: mcpproxy-windows-amd64.exe
             archive_format: zip
-          - os: ubuntu-latest
+          - os: windows-latest
             goos: windows
             goarch: arm64
-            cgo: "0"
+            cgo: "1"
             name: mcpproxy-windows-arm64.exe
             archive_format: zip
           - os: macos-15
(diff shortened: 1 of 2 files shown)


Dumbris:
Duplicate of #582 (which is the complete change incl. docs). Closing in favor of #582.

PR #617, opened 2026-06-06
fix(macos): tri-state duration fields in tray Settings + mandatory QA merge gate

Dumbris:
## Summary
Fixes the native macOS tray Settings → **Tool discovery & health checks** duration fields (spec 074), and adds a **mandatory QA merge gate** so this class of native-only bug can't ship again.

Reported via the cockpit (MCP-1214).

## The bugs (native tray only — Web UI was correct)
The form collapsed the `nil ↔ ""` tri-state (`nil`=default · `0s`=disabled · value) lossily:
1. **Save always disabled** — `.duration` validation rejected empty even when `optional`.
2. **Wrong placeholder** — hardcoded `"2m"` for every duration instead of the real defaults (30s / 5m).
3. **"2 unsaved changes" on first open** — `valuesEqual("", nil)` was `false`, so an absent optional field read dirty.

## Fix
- Validation respects `optional` (blank optional = inherit default).
- `ConfigField.placeholder`; discovery fields show real defaults **30s / 5m**.
- `valuesEqual` treats `nil`/`NSNull`/`""` as one blank class.
- `optionalStringBinding` / `optionalScalarStored`: a cleared field is stored as `NSNull` → PATCH sends JSON `null` (reset-to-default), never an unparseable `""`.
- Web parity: same `optional` fix in `fields.ts` + blank→null emit in `SettingField.vue`.
- Reads/writes only via `GET`/`PATCH /api/v1/config` — no direct config-file access.

## Mandatory QA merge gate
Closes the gap that let this ship (native tray is a separate hand-port; Web-UI-only QA never exercised it):
- `native-tests.yml` → required checks **`swift-test`** (native form logic, 253 green) + **`settings-parity`**.
- `check-settings-parity.py` — web↔native duration-field canary; it **caught a 2nd divergence** (`call_tool_timeout` native placeholder) → fixed here.
- `post-qa-gate-status.sh` — SHA-keyed `qa-gate` status the QATester posts at end-of-run.
- `mcpproxy-qa` skill: native tray = **mandatory surface**; interactive `cannot_verify` = **BLOCK**, not a low-risk pass.
- `docs/qa-merge-gate.md` — design + ordered branch-protection activation (keeps `--admin` escape hatch).

## Tests
- Swift: 9 new (`SettingsDiscoveryFieldsTests`), full suite 253 green w/ documented skips; the only skipped/failing are pre-existing & reproduce on clean `main`.
- Web: `settings-validation.spec.ts` (4) green.
- `settings-parity` green.

## Activation (post-merge)
Add `swift-test` + `settings-parity` to `main` required checks once this lands (a required-but-never-run check would block open PRs); stage `qa-gate` once QATester emits it. Commands in `docs/qa-merge-gate.md`.

Refs MCP-1214

### DIFF
--- a/.github/workflows/native-tests.yml
+++ b/.github/workflows/native-tests.yml
@@ -1,0 +1,71 @@
+# Native macOS tray test gate (MCP-1214).
+#
+# The macOS tray Settings form is a separate hand-port of the Web UI catalog,
+# so web-only QA cannot catch native-only behavior bugs (validation, dirty
+# detection, placeholders, bindings). This workflow makes that surface a
+# first-class, deterministic merge gate:
+#
+#   - swift-test       : runs the native unit tests (GUI-free logic), which
+#                        directly cover the spec-074 duration-field bugs.
+#   - settings-parity  : fails when the web (fields.ts) and native
+#                        (SettingsCatalog.swift) duration-field catalogs drift.
+#
+# Both job names are stable so they can be added to branch protection as
+# required status checks. See docs/qa-merge-gate.md.
+name: Native macOS Tests
+
+on:
+  pull_request:
+    paths:
+      - 'native/macos/**'
+      - 'frontend/src/views/settings/**'
+      - 'frontend/src/components/settings/**'
+      - 'scripts/check-settings-parity.py'
+      - '.github/workflows/native-tests.yml'
+  push:
+    branches: [main, next]
+    paths:
+      - 'native/macos/**'
+      - 'frontend/src/views/settings/**'
+      - 'frontend/src/components/settings/**'
+      - 'scripts/check-settings-parity.py'
+      - '.github/workflows/native-tests.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  swift-test:
+    name: swift-test
+    runs-on: macos-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4
+      - name: Show Swift toolchain
+        run: swift --version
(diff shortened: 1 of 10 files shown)


Dumbris:
## Added: required-safe rework of `native-tests.yml` (MCP-1219)

Pushed `fe9fb8e4` to this branch. This makes the QA-gate workflow **required-safe** before the `swift-test` / `settings-parity` contexts are added to `main` branch protection (MCP-1219).

**Why:** the previous `native-tests.yml` used a **top-level `paths:` filter**. On a PR that doesn't touch `native/macos/**` or the settings paths (e.g. a deps bump), the workflow never runs, so `swift-test`/`settings-parity` produce **no status**. The moment those are marked *required* on `main`, every such PR would be stuck on "Expected — Waiting" and **blocked forever**.

**Fix:** the workflow now runs on every PR (no top-level `paths:`). A `changes` job (`dorny/paths-filter`) detects native/settings changes, and `swift-test`/`settings-parity` are gated with a job-level `if:`. On non-native PRs they're **skipped** → the required context reports green; on native PRs they run for real. Job names are unchanged, so the context names stay stable.

Also updated `docs/qa-merge-gate.md` to document the design and require a non-native-PR verification before adding the contexts.

This PR still touches native paths, so `swift-test`/`settings-parity` run and pass here as before; the new `detect-native-changes` job runs too. No change to merge gating — this PR is still a **human Gate-3 merge**. Once it's on `main`, MCP-1219 verifies on a non-native PR, then adds the contexts per the doc.

PR #629, opened 2026-06-09
feat(macos): calm "Sign in" affordance for OAuth login-required servers (MCP-1822)

Dumbris:
## What & why

Parent epic: smart-mcp-proxy/mcpproxy-go MCP-1819 — *OAuth login-required = actionable Sign-in state, not MCPX_UNKNOWN_UNCLASSIFIED*. This is **T3 (macOS tray)**.

A server in the OAuth login-required state (`health.action == "login"`) read as a **hard error** in the tray. Pre-T1 the backend classifies that state as an `error`-severity `MCPX_UNKNOWN_UNCLASSIFIED` diagnostic, so the same server appeared in the red **"Fix issues"** section (`xmark.octagon` / "file a bug") *and* tinted the tray-icon badge red — while the only login affordance was a plain "Log In" menu item.

This renders it as a **calm, actionable Sign-in state** instead.

## Changes (all in `native/macos/`)

- `ServerStatus.isOAuthLoginRequired` — keyed off the stable, cross-surface `health.action == "login"` contract (the same one CLI/REST/Web-UI/TUI use).
- `AppState.serversWithDiagnostic` and `worstDiagnosticSeverity` **exclude** login-required servers → no red "Fix issues" entry, no red badge tint for a pure sign-in state. They remain in the calm **"Needs Attention"** group, which now shows the remediation verb (**"Sign in"**) explicitly so the affordance isn't buried.
- Rename **"Log In" → "Sign in"** across the tray (`HealthAction.login.label`, server submenu, ServerDetail header button, server-list context menu), mirroring the Web UI's calm tone.

### Contract / T1 relationship
Keyed off the **already-stable** `action == "login"` contract that exists on `main`, so this is independently correct and **forward-compatible** with T1's classifier change (whether T1 stops emitting a diagnostic for OAuth-pending or emits a calmer one, the tray no longer error-frames the sign-in state).

## Tests
`swift test` — new `AppStateOAuthSignInTests` (6) + updated `ModelsTests.testHealthActionLabels` all pass:
- login-required server excluded from Fix-issues / badge tint
- login-required server retained in Needs-Attention
- genuine non-OAuth error **still** surfaces in Fix-issues and tints the badge red
- `HealthAction.login.label == "Sign in"`

Pre-existing, environment-dependent failures unrelated to this change (`AutoStartTests.testFirstRunFlagRoundTrip` — UserDefaults pollution; `MergePatchEncodingTests` and `SSEParserTests` — Foundation-version canaries) reproduce on clean `origin/main` and are not touched here.

## Verification note
A live OAuth-pending upstream cannot be staged in this environment, so visual `mcpproxy-ui-test` capture of the exact login state isn't meaningful here; behavior is covered by the unit tests above against the `AppState`/`ServerStatus` logic that drives the menu.

Related #MCP-1819

### DIFF
--- a/native/macos/MCPProxy/MCPProxy/API/Models.swift
+++ b/native/macos/MCPProxy/MCPProxy/API/Models.swift
@@ -112,7 +112,7 @@ enum HealthAction: String, Codable, CaseIterable {
     /// Human-readable button label.
     var label: String {
         switch self {
-        case .login:      return "Log In"
+        case .login:      return "Sign in"
         case .restart:    return "Restart"
         case .enable:     return "Enable"
         case .approve:    return "Approve"
@@ -351,6 +351,15 @@ struct ServerStatus: Codable, Identifiable, Equatable {
     var hasAttentionDiagnostic: Bool {
         guard let d = diagnostic, !(d.code).isEmpty else { return false }
         return d.severity == "warn" || d.severity == "error"
+    }
+
+    /// True when the server is in the OAuth login-required state (MCP-1819/T3).
+    /// `health.action == "login"` is the stable, cross-surface contract that
+    /// CLI/REST/Web-UI/tray all key off. In this state the server needs a calm,
+    /// actionable "Sign in" affordance — NOT hard-error framing — even when the
+    /// backend also attaches an error-severity diagnostic for the failed connect.
+    var isOAuthLoginRequired: Bool {
+        health?.action == "login"
     }
 
     /// Number of tools awaiting approval (pending + changed), or 0 if quarantine stats are absent.
--- a/native/macos/MCPProxy/MCPProxy/API/Models.swift
+++ b/native/macos/MCPProxy/MCPProxy/API/Models.swift
@@ -390,6 +390,14 @@ struct ServerStatus: Codable, Identifiable, Equatable {
         default: return connected ? .systemGreen : .systemGray
         }
     }
+
+    /// AppKit tray-menu dot color (MCP-1822). The OAuth login-required state is
+    /// calm and actionable, not a failure: it gets the system accent tint rather
+    /// than the red/error `statusNSColor` that previously read as a hard error.
+    /// Genuine errors (any non-login state) keep `statusNSColor`.
+    var menuStatusNSColor: NSColor {
+        isOAuthLoginRequired ? .controlAccentColor : statusNSColor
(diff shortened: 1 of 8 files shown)


Dumbris (commented):
## Code Review: MCP-1834 — macOS tray Sign-in affordance

**Overall:** Core logic is sound and test coverage is thorough. One structural issue before merge: `isOAuthLoginRequired` was added as the "stable, cross-surface contract" but most call sites still use the raw check.

### Required fix

**Inconsistent use of `isOAuthLoginRequired`**

AppState.swift correctly adopts the new computed property. But three call sites still use the raw `health?.action == "login"` check:

- `TrayMenu.swift:256`
- `ServerDetailView.swift:144`
- `ServersView.swift:507`

Change all three to `server.isOAuthLoginRequired` so the abstraction is actually used and a future contract change only needs to happen in one place.

### Observation (non-blocking)

The new label block in TrayMenu.swift shows the action label for any health action, not only login. A restart-needed server would also show "Restart" as secondary text. Confirm this is intentional or add an `isOAuthLoginRequired` guard.

### What's good

- `isOAuthLoginRequired` computed property — correct abstraction
- AppState exclusion logic (Fix Issues + badge tint) — correct and well-commented
- `AppStateOAuthSignInTests.swift` — comprehensive, covers all three AppState facets
- Consistent "Sign in" rename across all surfaces

**Fix the three raw `health?.action == "login"` checks → this is ready to merge.**

Dumbris:
Thanks for the review (MCP-1834) — both points addressed in `e292ff3a`:

**Required fix (done):** the three call sites now use `server.isOAuthLoginRequired` instead of raw `health?.action == "login"` — TrayMenu submenu, ServerDetail header button, and ServersView context menu. The stable contract now has a single definition and consistent call sites.

**Non-blocking observation (guarded):** the Needs-Attention trailing remediation label is now scoped to the login state only (`if server.isOAuthLoginRequired`). Surfacing verbs for *all* attention actions was out of scope for this login-focused issue, so other actions keep their prior icon-only row. Happy to broaden that to all actions in a follow-up if desired.

`swift test`: AppStateOAuthSignInTests (6) + label test remain green; behavior-identical for the login path. Re-dispatched QA (SHA-pinned to the new head) since the push invalidates the prior `qa-gate`.

PR #643, opened 2026-06-12
fix(scanner): resolve source for runtime-added Docker-isolated stdio servers (MCP-2123 Defect A)

Dumbris:
## Summary

Fixes **Defect A** of MCP-2123: security scans of a Docker-isolated `stdio` server added at runtime from the official registry (slash/dot names like `com.pulsemcp/google-flights`) reported **"No Source Available"** (`source_method=none`), exported no `tools.json`, and skipped Pass 2 with *"No server info available"*.

Two independent root causes, both confirmed by static trace from the issue's evidence:

### 1. Stale config snapshot (primary — matches all reported evidence)
`configServerInfoProvider` iterated a `*config.Config` captured at **server boot**. `runtime.Config()` returns a fresh immutable snapshot on every reload, so a server **added at runtime** (exactly the registry case) was absent from the boot snapshot → `GetServerInfo` returned *not found* → the scanner ran with **no `ServerInfo`**:
- source resolution never ran → `source_method` stayed `none`;
- tool definitions were never exported → `/scan/source/tools.json` missing → `tpa-descriptions` / `mcp-scan` failed;
- Pass 2 got `serverInfo == nil` → *"No server info available for Pass 2, skipping"*.

The provider now resolves against the **live** config (`runtime.Config`), falling back to the boot snapshot when no accessor is wired.

### 2. Container-name sanitizer mismatch (secondary)
The scanner's `docker ps --filter name=mcpproxy-<sanitized>-` lookup mapped `.`→`-`, while the launcher (`internal/upstream/core`) **preserves** `.` (Docker allows it). So for any dotted name the filter prefix never matched the real container (`mcpproxy-com.pulsemcp-google-flights-umqu`) and `docker_extract` silently fell through. Both sides now share **`internal/dockernaming.SanitizeServerName`** so the rule can't drift again.

With (1) fixed the scan resolves `ServerInfo` and (worst case) exports `tools.json` via the existing `tool_definitions_only` path; with (2) fixed `docker_extract` matches the container and scans the real package source.

## Tests
- `internal/dockernaming`: `TestSanitizeServerName` — locks the `com.pulsemcp/google-flights` → `com.pulsemcp-google-flights` case plus full parity with the launcher's prior rules.
- `internal/server`: `TestConfigServerInfoProvider_UsesLiveConfig` / `_FallsBackToSnapshot` — regression guard for the stale-snapshot bug.
- `internal/upstream/core` sanitizer tests still pass (behavior preserved via delegation).

## Verification
- `go build ./...` ✅
- `go test -race ./internal/dockernaming/... ./internal/security/scanner/... ./internal/upstream/core/... ./internal/server/...` ✅
- `./scripts/run-linter.sh` → `0 issues` ✅
- `gofmt -l` clean, `go vet` clean ✅

## Scope
- **In scope:** Defect A (backend), `internal/` only.
- **Defect B** (frontend: scan-report "View Full Report" link not `encodeURIComponent`-ing the scan id) → separate FrontendEngineer child issue.
- **Infra follow-up** (ramparts arm64 GLIBC, trivy DB-download timeout) → noted, out of scope.

No public API/CLI/config contract change; no docs change required (restores the behavior `source_method` docs already describe).

Related #MCP-2123

### DIFF
--- a/internal/dockernaming/naming.go
+++ b/internal/dockernaming/naming.go
@@ -1,0 +1,64 @@
+// Package dockernaming centralises the rule for turning an MCPProxy server name
+// into the sanitized token used inside a Docker container name
+// (mcpproxy-<sanitized>-<suffix>).
+//
+// It exists so the component that NAMES a container at launch
+// (internal/upstream/core) and the component that LOOKS UP a running container
+// by name (internal/security/scanner) share one source of truth. They used to
+// carry independent sanitizers that disagreed on '.': the launcher preserved it
+// (Docker allows '.') while the scanner mapped it to '-'. For official-registry
+// servers — whose names are a dotted namespace plus a slash, e.g.
+// "com.pulsemcp/google-flights" — that mismatch meant the scanner's
+// `docker ps --filter name=mcpproxy-<sanitized>-` prefix never matched the real
+// container, so source extraction silently fell through to "No Source Available"
+// (MCP-2123). Keeping the rule in one leaf package makes that drift impossible.
+package dockernaming
+
+import (
+	"regexp"
+	"strings"
+)
+
+var (
+	invalidContainerChars = regexp.MustCompile(`[^a-zA-Z0-9_.-]+`)
+	leadingAlphanumeric   = regexp.MustCompile(`^[a-zA-Z0-9]`)
+)
+
+// maxSanitizedLen bounds the sanitized token so the full container name
+// (mcpproxy- prefix + token + - + 4-char suffix) stays well under Docker's
+// 253-char limit.
+const maxSanitizedLen = 200
+
+// SanitizeServerName converts a server name into a valid Docker container-name
+// token. Docker container names may contain [a-zA-Z0-9][a-zA-Z0-9_.-]*, so this
+// preserves letters, digits, '_', '.', and '-' and replaces any other run of
+// characters with a single '-'. It then collapses consecutive hyphens, ensures
+// the result starts with an alphanumeric character, trims trailing '-'/'.', and
+// truncates to a safe length.
+func SanitizeServerName(name string) string {
+	sanitized := invalidContainerChars.ReplaceAllString(name, "-")
+
+	for strings.Contains(sanitized, "--") {
+		sanitized = strings.ReplaceAll(sanitized, "--", "-")
+	}
+
+	if sanitized != "" && !leadingAlphanumeric.MatchString(sanitized) {
+		sanitized = "server-" + sanitized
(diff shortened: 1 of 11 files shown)


Dumbris:
## 🔎 Verification (CEO cockpit) — CHANGES REQUESTED

Built this branch **merged with #644** and ran a real end-to-end scan of a runtime-added, Docker-isolated `com.pulsemcp/google-flights` (npx). The two fixes here are **correct**, but they expose **two more backend defects** that must land before slash-named (official-registry) servers can actually be scanned. Tracked + patch attached in **MCP-2149** (assigned BackendEngineer, inherits this branch's workspace).

**What this PR fixes (confirmed):** stale config snapshot → ServerInfo found; sanitizer unified → container now matches `mcpproxy-com.pulsemcp-google-flights-…`.

**Newly-exposed, blocking:**
1. `os.MkdirTemp` rejects slash-named servers → extraction fails with `pattern contains path separator` (sites: `source_resolver.go:355,814`, `service.go:682`). Fix: `dockernaming.SanitizeServerName(serverName)`.
2. `handleGetScanReportByJobID` reads raw `chi.URLParam(r,"jobId")` (no decode) → report endpoint 404s for slash jobIds even after #644. Fix: `decodePathParam(...)` at `security_scanner.go:666`.

**With both applied (verified):** `source_method=docker_extract`, 23 files extracted, tools_exported=3, scanner completed; report endpoint → HTTP 200; Web UI shows "Source extracted from container (23 files)" and the report page renders (no 404).

Recommend folding MCP-2149's patch into this branch so it ships as one PR; QA will re-bless the new head SHA.

Dumbris:
Pushed follow-up commit `d610cba1` addressing two defects the sanitizer fix exposed for slash-named official-registry servers (verified end-to-end by CEO on `com.pulsemcp/google-flights`):

1. **`os.MkdirTemp` rejected slash names** — sanitized all three extraction temp-dir patterns (`mcpproxy-scan-%s-`, `-scan-full-%s-`, `-scan-tools-%s-`) via `dockernaming.SanitizeServerName`.
2. **`handleGetScanReportByJobID` raw `jobId`** — now `decodePathParam(...)` so the scan-report endpoint stops 404ing (scan ids embed the slash-containing server name).

Verified result: `source_method=docker_extract`, 23 files extracted, `tools_exported=3`, report endpoint HTTP 200. New regression tests: `TestScanTempDirPatternHandlesSlashNames`, `TestDecodePathParamScanJobID`. Local `-race` + `run-linter.sh` (0 issues) green.

PR #657, opened 2026-06-14
fix(logs): isolate non-default data-dir logs to <data-dir>/logs (MCP-2255)

Dumbris:
## Summary

When `mcpproxy serve` runs with a non-default `--data-dir` and no explicit `--log-dir`,
co-locate logs under `<data-dir>/logs` instead of the shared OS-standard location
(`~/Library/Logs/mcpproxy` on macOS, `~/.local/share/mcpproxy/logs` on Linux).

### Why

Every process that runs `mcpproxy serve` — Go integration tests, e2e scripts,
FE/QA Playwright harnesses, and the real production core — was appending to the
**same shared `main.log`**. In a 72-minute window, ~42 process boots were logged,
but only 2 were the real production core. The other 40 were test/harness runs.
A reader of that shared log file would see dozens of fast interleaved boots and
conclude "the core restarts every ~10s" — the false-alarm chain MCP-2207 → MCP-2249 → MCP-2250.

### Resolution order

`--log-dir` flag → config `logging.log_dir` → `<data-dir>/logs` (non-default data dir only)
→ OS-standard location.

### Changes

| File | Change |
|------|--------|
| `cmd/mcpproxy/logdir.go` | **NEW** — `resolveServeLogDir()` + helpers |
| `cmd/mcpproxy/logdir_test.go` | **NEW** — table-driven tests for all precedence cases |
| `cmd/mcpproxy/main.go` | Wire `resolveServeLogDir()` into `runServer()` |
| `docs/logging.md` | Document the custom data-dir log location |
| `internal/logs/e2e_test.go` | Assert logs go to `<data-dir>/logs`; verify **no leak** into shared OS dir |

### Testing

- `make build` — passes
- `go test ./cmd/mcpproxy/... -race -count=1` — passes
- `go test ./internal/logs/... -race -count=1` — passes
- `./scripts/run-linter.sh` — 0 issues

### Gates

- **Do not merge** — human pre-merge gate (Gate 3)

Related: MCP-2255

PR #660, opened 2026-06-14
fix(logs): isolate non-default data-dir logs to <data-dir>/logs

Dumbris:
## Summary

Route a **non-default** data-dir's logs to `<data-dir>/logs` instead of the shared OS-standard location (`~/Library/Logs/mcpproxy/` on macOS). This keeps Go test runs, e2e scripts, and FE/QA Playwright harnesses from polluting the production `main.log`.

## Root cause

Forensics in MCP-2249: the phantom "core restarts every ~10s" was multiple test instances interleaving re-init lines in one shared log file. Every `mcpproxy serve` with a custom `--data-dir` wrote to the same `~/Library/Logs/mcpproxy/main.log` because the log dir was derived purely from `$HOME`.

## Design

**Precedence** (highest to lowest):
1. `--log-dir` flag — always wins
2. `logging.log_dir` in config
3. `<data-dir>/logs` — only when data-dir is **not** the default (`~/.mcpproxy`)
4. OS-standard location (unchanged for default data dir)

Default data dir (`~/.mcpproxy`) is unaffected — it still logs to the OS-standard location so the tray and documented paths keep working.

## Files changed

- `cmd/mcpproxy/logdir.go` — new: `resolveServeLogDir()`, `defaultDataDirPath()`, `sameDir()`
- `cmd/mcpproxy/logdir_test.go` — new: 7 table-driven tests covering every branch
- `cmd/mcpproxy/main.go` — replaces inline `if cmdLogDir != ""` override with `resolveServeLogDir()`
- `docs/logging.md` — documents the new "Custom data directory" log location
- `internal/logs/e2e_test.go` — updates e2e test to assert co-located path; adds negative assertion that OS-standard path is NOT written

## Verification

- `go test ./cmd/mcpproxy/... -race -run TestResolveServeLogDir` — 7/7 PASS
- `make build` — compiles clean

Related #2255

### DIFF
--- a/cmd/mcpproxy/logdir.go
+++ b/cmd/mcpproxy/logdir.go
@@ -1,0 +1,65 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/smart-mcp-proxy/mcpproxy-go/internal/config"
+)
+
+// defaultDataDirPath returns the default data directory (<home>/.mcpproxy),
+// mirroring internal/config's default resolution. It is used to decide whether
+// the resolved data dir is "non-default" for log co-location. On failure to
+// resolve the home dir it returns config.DefaultDataDir, which will simply not
+// match any absolute data dir and leave logs at the OS-standard location.
+func defaultDataDirPath() string {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return config.DefaultDataDir
+	}
+	return filepath.Join(home, config.DefaultDataDir)
+}
+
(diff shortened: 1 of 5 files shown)


Dumbris:
Closing as a duplicate of #657. Both PRs carry byte-identical changes for MCP-2255 (log isolation for non-default data dirs); #657 is canonical (opened earlier, CI fully green, title/body reference MCP-2255). Consolidating on #657 to avoid two open PRs for the same fix.

PR #674, opened 2026-06-15
fix(scanner): ramparts scan fails closed on garbled tools.json or scan error

Dumbris:
## Problem (MCP-2443, [SEC] HIGH)

The Ramparts docker scanner **failed OPEN**: a broken/garbled input or a ramparts failure was reported as a *clean, zero-finding* scan, silently defeating the security gate.

Two root causes:
1. `mcp-replay.py` returned `[]` tools when `/scan/source/tools.json` was missing / empty / garbled / wrong-shape → Ramparts completed a handshake against an empty tool list and emitted a "clean" report.
2. `entrypoint.sh` masked the ramparts exit code (`|| true`) and accepted any non-empty file (`[ -s REPORT ]`) as success → a failed scan that wrote a non-empty **error** JSON was treated as a successful report.

Compounding it: the Go runner **ignores the container exit code** (`docker.go` `RunScanner` returns `nil` for a non-zero exit) and marks a scan failed only when *no readable report exists*. So `exit 1` alone is not enough — the engine reads whatever report file is present.

## Fix — fail CLOSED

- **`mcp-replay.py`**: `load_tools()` now raises `ToolsLoadError` and the shim exits non-zero **serving no frames** when `tools.json` is unreadable / empty / invalid JSON / wrong-shape / yields zero usable tools. Ramparts then sees a dead endpoint instead of an empty tool list. A partially-garbled file with ≥1 real tool still scans.
- **`entrypoint.sh`**: captures the ramparts exit code, then validates the output is a **genuine Ramparts report** (top-level `security_issues` object + `yara_results` array — the same shape `isRampartsOutput()` probes). On any failure it **deletes the report** so the engine finds no output and records the scanner as *failed*, instead of parsing a stale/error payload as 0 findings. A non-zero ramparts exit **with a valid report** is kept (findings/offline-LLM signal that way — discarding it would lose real detections).

## Acceptance

| Criterion | Covered by |
|---|---|
| Garbled/empty `tools.json` → non-zero exit, scan failed (not clean) | `mcp-replay_test.py`: `test_{empty_file,garbled_json,wrong_shape,empty_tool_list,missing_tools_file}_fails_closed` |
| Non-zero ramparts exit / error payload → propagated as scan failure | `entrypoint_test.sh`: `error_payload_nonzero_fails`, `error_payload_rc0_fails`, `empty_report_fails`, `garbled_report_fails` |
| Valid `tools.json` with a real finding still scans + reports | `mcp-replay_test.py`: `test_partial_garble_with_real_tool_still_scans`; `entrypoint_test.sh`: `valid_*_kept` |

## Verification (local)

```
$ python3 docker/scanners/ramparts/mcp-replay_test.py   →  11/11 passed
$ sh docker/scanners/ramparts/entrypoint_test.sh        →  7 passed, 0 failed
$ shellcheck docker/scanners/ramparts/entrypoint*.sh    →  clean (exit 0)
```

Docs updated: `docs/features/scanner-images.md` (fail-closed section).

## Notes
- No Go changes; behavior is enforced in the scanner container assets. The Dockerfile `COPY`s only `entrypoint.sh`/`mcp-replay.py`, so the new `*_test.sh`/`*_test.py` files don't enter the image.
- Follow-up (Release lane): wire `mcp-replay_test.py` + `entrypoint_test.sh` into `scanner-images.yml` CI (they currently run locally only).

Related MCP-2443

### DIFF
--- a/docker/scanners/ramparts/entrypoint.sh
+++ b/docker/scanners/ramparts/entrypoint.sh
@@ -62,10 +62,12 @@ if [ ! -s "$REPORT" ]; then
 fi
 
 # Validate $REPORT is a GENUINE Ramparts report, not an error payload. A real
-# report carries a top-level "security_issues" object and a "yara_results"
-# array (mirrors the engine's isRampartsOutput() probe). An error/garbled JSON
-# lacks these; without this gate it would parse as 0 findings = a spurious clean
-# scan (fail-open). python3 is already in the image for the replay shim.
+# report carries a top-level "security_issues" OBJECT (with the tool/prompt/
+# resource issue arrays) and a "yara_results" ARRAY. We TYPE-check, not just
+# presence-check: an error payload such as {"security_issues":"scan failed"}
+# has the key present but the wrong type and must fail CLOSED — a presence-only
+# gate would wave it through as a spurious clean scan (MCP-2443 re-review).
+# python3 is already in the image for the replay shim.
 if ! python3 - "$REPORT" <<'PY'
 import json, sys
 try:
@@ -75,9 +77,10 @@ except (OSError, ValueError) as exc:
     print(f"report is not valid JSON: {exc}", file=sys.stderr)
     sys.exit(1)
 if (not isinstance(data, dict)
-        or data.get("security_issues") is None
+        or not isinstance(data.get("security_issues"), dict)
         or not isinstance(data.get("yara_results"), list)):
-    print("report lacks security_issues/yara_results (error payload or wrong shape)", file=sys.stderr)
+    print("report is not a Ramparts report: security_issues must be an object "
+          "and yara_results an array (error payload or wrong shape)", file=sys.stderr)
     sys.exit(1)
 sys.exit(0)
 PY
--- a/docker/scanners/ramparts/entrypoint.sh
+++ b/docker/scanners/ramparts/entrypoint.sh
@@ -24,19 +24,69 @@
 # them at /scan and WORKDIR is /scan (see Dockerfile).
 set -u
 
-REPORT=/scan/report/results.json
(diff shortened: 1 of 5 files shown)


Dumbris:
**codex review (gpt-5.5) → REQUEST_CHANGES** (independently re-verified)

Blocking bug: the fail-closed gate in `entrypoint.sh` checks `data.get('security_issues') is None` (**presence**, not **type**). An error payload like `{"security_issues":"scan failed","yara_results":[]}` (string, not object) passes the gate and is kept as a valid clean report — contradicting the PR's own doc ("a top-level `security_issues` **object**"). 

Fix: require `isinstance(data.get('security_issues'), dict)` (and similarly type-check the other expected-object fields). Add a test for the malformed-string payload.

Dumbris:
## Re-review fix pushed (type-confusion gap closed)

Thanks — confirmed real. The fail-closed gate was presence-only (`security_issues is None`), so `{"security_issues":"scan failed","yara_results":[]}` passed as clean.

**Fix (`ca59f70c`):** `entrypoint.sh` now **type-checks** the report shape — `security_issues` must be an object (`dict`) and `yara_results` an array (`list`), matching the real Ramparts schema (`{security_issues{tool_issues,...}, yara_results[]}`). Anything else fails CLOSED (report deleted → scanner marked failed).

**TDD:** `entrypoint_test.sh` adds `string_security_issues_fails` and `string_yara_results_fails` — both assert the scan is marked failed (report removed), not clean.

Verification: entrypoint **9/9**, shim **11/11**, shellcheck clean. Pushed to the same branch; CI + review re-trigger. QA re-dispatched at head `ca59f70c`.

PR #684, opened 2026-06-15
feat(telemetry): one-time opt-out beacon + banner Settings deep-link (MCP-2482)

Dumbris:
## Summary
Implements **MCP-2482** — a telemetry consent banner Settings deep-link (Part A) and a one-time, server-side opt-out beacon (Part B). Reuses the MCP-2477 resolved-state convention (`IsTelemetryEnabled` — `nil` means enabled).

### Part B — server-side opt-out beacon (`internal/telemetry`)
- `TelemetryDisableTransition(prior, next)` — detects the enabled→disabled flip by comparing **resolved** telemetry state.
- `SendOptOutBeacon` — posts `{"event":"telemetry_disabled","anonymous_id":"…"}` to the **existing `/heartbeat` ingest** (no new endpoint). Carries **only** the anonymous install ID — **no usage payload**. Runs through the same `ScanForPII` guard.
- `Service.NotifyConfigChanged(newCfg)` — on a flip: latches `optedOut` (stops all further heartbeats) then fires **exactly one** best-effort, fire-and-forget beacon (5s timeout). A send failure **still disables**. `disabled→disabled` / reload-while-disabled / env-disabled emit nothing. Re-enable clears the latch (exactly once per flip).
- **One implementation, three covered paths:** wired into `runtime.ApplyConfig` (REST → web UI + macOS app), `runtime.ReloadConfiguration` (disk reload), and `mcpproxy telemetry disable` (CLI — there is no fsnotify watcher, so the CLI fires its own beacon).

### Part A — banner deep-link + disclosure (Vue)
- `TelemetryBanner` gains a **"Manage in Settings"** link → `/settings?focus=telemetry.enabled`, plus a transparency note.
- `Settings.vue` `focusField()` resolves a field key to its tab, opens the enclosing accordion, scrolls to + highlights the row.
- Disclosure line near the toggle: *"Disabling sends a single anonymous opt-out signal, then stops all telemetry."* (field help + confirm copy).

## Tests (TDD)
- `internal/telemetry/optout_test.go`: transition table; beacon payload shape (path/method, event+anon ID, **asserts no usage fields**); fires exactly once on disable; nothing when already disabled; **send failure still disables**; **no further heartbeats after opt-out**.

## Verification
- `go build ./...` + `-tags server` ✅ · `go test ./internal/telemetry ./internal/runtime ./cmd/mcpproxy -race` ✅
- `golangci-lint v2.5.0` (CI config) → 0 issues · `gofmt` clean · `vue-tsc --noEmit` clean · `make build` ✅
- `scripts/test-api-e2e.sh` → 65/65 ✅
- **Real CLI smoke:** `telemetry disable` against a capture server → exactly 1 beacon `{"event":"telemetry_disabled","anonymous_id":"smoke-anon-001"}` to `/v1/heartbeat`; second `disable` → 0.

## Docs
`docs/features/telemetry.md` documents the one-time opt-out signal.

Related #MCP-2482


Dumbris:
**codex review (gpt-5.5, in-checkout) → REQUEST_CHANGES.** Beacon payload shape and the banner deep-link + disclosure are good. Three blocking issues (2× P1 privacy-relevant):

1. **[P1] Env-resolved disabled state not respected.** `NotifyConfigChanged`/CLI gate on `Config.IsTelemetryEnabled()` (`config.go:1716`), which only honors `MCPPROXY_TELEMETRY=false` — but real telemetry startup ALSO disables on `DO_NOT_TRACK`/`CI` (`telemetry.go:372`, `env_overrides.go:26`). So `DO_NOT_TRACK=1` / `CI=true` can emit an opt-out beacon when telemetry was never actually enabled. Resolve the transition through the SAME effective-enabled check used at startup (incl. env overrides), not just the config bool.
2. **[P1] Usage heartbeat can post AFTER opt-out (race).** `sendHeartbeat` checks `optedOut` only once at entry (`telemetry.go:431`); a heartbeat that passes the check just before `NotifyConfigChanged` flips the latch (`optout.go:181`) still sends the full usage payload. Re-check the latch immediately before transmit (or hold a lock across the flip) so no usage data leaves after opt-out.
3. **[P2] CLI disable is a separate synchronous sender.** After saving disabled config it blocks in `SendOptOutBeacon` under a 5s timeout (`telemetry_cmd.go:292`) and bypasses the runtime semver/dev-build guard. Route CLI disable through the same server-side fire-and-forget beacon path with the same guards.

PR #685, opened 2026-06-15
feat(ui): prefill MCP instructions default + Reset-to-default button (MCP-2484)

Dumbris:
## What

- **Prefill** the `instructions` textarea with the built-in default (from `/api/v1/status → default_instructions`) when the user hasn't saved a custom value yet (fresh install / blank field)
- **Reset to default button** — compact inline button next to the textarea that re-populates the field with the live backend default; appears only after the default resolves

## Why

Fresh installs showed an empty textarea with no hint of what the built-in instructions look like. Users had no way to restore the default after clearing it.

### `fields.ts`
- Added `resetDefault?: string` to the `SettingField` interface
- Exported `isBlankInstructions(v)` helper — treats `null`/`undefined`/empty/whitespace as blank (eligible for prefill)

### `SettingField.vue`
- In the `textarea` control block: render a compact `↩ Reset to default` button when `field.resetDefault` is set; hidden until the async default resolves (no `undefined` flash)
- Click emits `resetDefault` through the normal `update:modelValue` path so the section tracks it dirty and Save persists it

### `Settings.vue`
- `advancedAccordions` computed: injects `placeholder` and `resetDefault` from `defaultInstructions` into the instructions field
- `maybePrefillInstructions()`: uses `isBlankInstructions` so whitespace-only saved values also get prefilled
- Dual-trigger: called from both `loadConfig()` completion and `loadDefaultInstructions()` completion + a watcher on `defaultInstructions` — whichever resolves last triggers prefill

### Tests
- 9 new unit tests in `frontend/tests/unit/instructions-prefill.spec.ts`
- All 186 frontend unit tests pass

## Guard: never overwrite saved value

`maybePrefillInstructions` only runs when `isBlankInstructions(state.working.instructions)` — a saved custom value is never touched.

Closes MCP-2484

### DIFF
--- a/frontend/src/components/settings/SettingField.vue
+++ b/frontend/src/components/settings/SettingField.vue
@@ -137,6 +137,14 @@
 
       <!-- textarea (multi-line text, e.g. MCP server instructions) -->
       <div v-else-if="field.control === 'textarea'" class="flex flex-col items-stretch w-full sm:w-96">
+        <div v-if="field.resetDefault != null" class="flex justify-end mb-1">
+          <button
+            type="button"
+            class="btn btn-ghost btn-xs text-base-content/50 hover:text-base-content"
+            :data-test="`setting-reset-${field.key}`"
+            @click="emitText(field.resetDefault ?? '')"
+          >↩ Reset to default</button>
+        </div>
         <textarea
           class="textarea textarea-bordered textarea-sm w-full font-mono leading-snug"
           :class="{ 'textarea-error': validationError }"
--- a/frontend/src/components/settings/SettingsSection.vue
+++ b/frontend/src/components/settings/SettingsSection.vue
@@ -5,7 +5,7 @@
       :key="f.key"
       :field="f"
       :model-value="getPath(working, f.key)"
-      :dirty="!!dirty[f.key]"
+      :dirty="isFieldDirty(f.key)"
       @update:model-value="onChange(f, $event)"
     />
     <p v-if="!fields.length" class="text-sm text-base-content/50 py-4">No settings match your search.</p>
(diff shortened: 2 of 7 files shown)


Dumbris:
**codex review (gpt-5.5, in-checkout) → REQUEST_CHANGES.** Async prefill, non-overwrite, and Reset (live default) all look correct — but the core "Save persists the text" requirement is not met:

- `Settings.vue:234` prefills `state.working.instructions` directly, but `SettingsSection.vue:116` only marks a field dirty via `onChange`, and `SettingsSection.vue:187` PATCHes only `dirtyKeys`.
- So on the **main prefill path**, the default is displayed but **not saveable/persisted** unless the user edits the field or clicks Reset. A user who opens Settings and hits Save expecting to keep the (now-visible) default gets nothing persisted.

Fix: when prefilling the default into the textarea, also mark `instructions` dirty (route the prefill through the same onChange path, or add it to dirtyKeys) so Save persists it. (Reset already works because it goes through onChange.) Add a Playwright assertion: fresh install → open Settings → Save without editing → reload → instructions persisted.

Dumbris:
## Codex REQUEST_CHANGES addressed — pushed to this branch (e147de80)

**Finding:** `maybePrefillInstructions()` wrote the default straight onto `state.working.instructions`, but `SettingsSection` only PATCHed its `onChange`-tracked dirty keys → Save-without-editing persisted nothing.

**Fix (frontend-only, `SettingsSection.vue`):** derive dirtiness from `working !== original` (unioned with the explicit `dirty` ref), so a value set outside a control — the prefill — is treated as dirty and included in `buildPartial`/PATCH. Reset already worked (goes through `onChange`); user-edit, discard, and post-save clearing semantics are unchanged.

**Verification**
- New component test `settings-section-prefill-dirty.spec.ts`: prefilled value (working≠original) → Save button enabled → `patchConfig` called with `{instructions: …}`; dirty clears after save; equal working/original stays non-dirty.
- New Playwright `specs/060-settings-page/verification/instructions-prefill-persist.spec.ts`, run against a fresh mcpproxy: **fresh install → open Settings → Save without editing → reload → instructions persisted** ✓, and Reset-to-default repopulates ✓.
- All **185** frontend unit tests pass; `vue-tsc` clean; `make build` green.

Acceptance criteria from the review (prefill-then-Save persists; existing edit/Reset/no-overwrite scenarios still pass) are all covered.

PR #686, opened 2026-06-15
feat(ui): prefill MCP instructions default + Reset-to-default button (MCP-2484)

Dumbris:
## Summary
Frontend-only UX fix for **Settings → "MCP server instructions"** (MCP-2484).

Previously the textarea's editable VALUE was the config `instructions` field (empty on a fresh install) and the built-in default was shown only as a **placeholder** — a user wanting to tweak/append the default had to retype the whole thing.

### Changes
1. **Prefill the editable value** with the live `default_instructions` (from `/api/v1/status`) when the saved `instructions` is empty. Handles either load order (config vs. default request) plus a `watch` so a late-arriving default still prefills. **Never overwrites a non-empty saved value.**
2. **Compact "Reset to default" button** next to the instructions textarea (shown only once the default has loaded) that repopulates the field from the live default. It flows through the normal change path, so the section marks it dirty and **Save persists it as `instructions`**.

Prefilling-then-saving "freezes" the current default into `config.instructions` (intended per the request); the Reset button is the re-sync path. No backend change — the backend already exposes the resolved default.

### Verification
- `npx vitest run` → **175 passed** (incl. 4 new `SettingField` reset-button tests + existing instructions tests).
- `npm run type-check` (vue-tsc) → clean.
- `make build` → frontend embed + Go binaries green.

### Test plan (added unit coverage)
- Reset button renders only when `resetDefault` is injected (async default loaded); hidden otherwise.
- Clicking Reset emits the default text through the change path.
- Existing empty-means-default / whitespace-collapse behavior preserved.

> ⚠️ Note for reviewers/CEO: a concurrent stuck agent run (session `443647fa`, no Paperclip checkout) was re-saving the same implementation into the original `mcp-2484-instructions-prefill` worktree without ever committing. This PR was produced in an isolated clean worktree by the legitimate Paperclip checkout owner to avoid the race. If that run later opens a duplicate PR, close one.

Related #MCP-2484

### DIFF
--- a/frontend/src/components/settings/SettingField.vue
+++ b/frontend/src/components/settings/SettingField.vue
@@ -137,6 +137,14 @@
 
       <!-- textarea (multi-line text, e.g. MCP server instructions) -->
       <div v-else-if="field.control === 'textarea'" class="flex flex-col items-stretch w-full sm:w-96">
+        <div v-if="field.resetDefault != null" class="flex justify-end mb-1">
+          <button
+            type="button"
+            class="btn btn-ghost btn-xs text-base-content/50 hover:text-base-content"
+            :data-test="`setting-reset-${field.key}`"
+            @click="emitText(field.resetDefault ?? '')"
+          >↩ Reset to default</button>
+        </div>
         <textarea
           class="textarea textarea-bordered textarea-sm w-full font-mono leading-snug"
           :class="{ 'textarea-error': validationError }"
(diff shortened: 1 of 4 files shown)


Dumbris:
Closing as a duplicate of #685 (same MCP-2484 frontend change, created ~5 min earlier and board-announced). #685 is the canonical PR: fuller implementation (whitespace-aware `isBlankInstructions()`, dual-trigger prefill + watcher) with CI nearly all green. This branch (`fix/mcp-2484-instructions-prefill`) was opened by a sibling run on the same issue; consolidating on #685 to avoid divergence.

PR #688, opened 2026-06-15
feat(broker): CredentialResolver — per-user-only ordering + policy seam (spec 074, MCP-1039)

Dumbris:
## Spec 074 — T6: `CredentialResolver`

The per-user credential resolver that decides which brokered credential to inject on a proxied request for the **server edition** upstream token broker (`internal/serveredition/broker/`). Builds on **T3** (`CredentialStore` + IdP subject token, [#601](https://github.com/smart-mcp-proxy/mcpproxy-go/pull/601)), **T4** (`TokenExchanger`, [#600](https://github.com/smart-mcp-proxy/mcpproxy-go/pull/600)), **T5** (`OAuthConnector`, [#602](https://github.com/smart-mcp-proxy/mcpproxy-go/pull/602)) — all merged to `main`.

### Per-user-only ordering (FR-013 / FR-014)
`Resolve(ctx, userID, server)` applies a strict ordering with **no shared/static fallback**:
1. valid cached per-user credential (refreshed if within the near-expiry window);
2. else token-exchange / Entra OBO from the user's stored IdP subject token;
3. else, for `oauth_connect` upstreams the user has not connected → an actionable `NotConnectedError` carrying the **connect URL**;
4. else `ErrNoCredential`.

### Design
- **Single-flight** coalescing per `(user, server)` (`golang.org/x/sync/singleflight`) so concurrent requests don't trigger duplicate upstream token flows.
- **Policy-decision seam** (`PolicyHook`, FR-015) evaluated per call *before* the credential is returned. Ships with an allow-all default — no policy engine yet, the seam is the deliverable.
- Unauthenticated callers (`ErrUnauthenticated`) and a disabled store (`ErrStoreDisabled`) are rejected up front.
- `Exchanger` / `Connector` / `ConnectorProvider` interfaces decouple the resolver from the concrete `TokenExchanger` and `OAuthConnector` (compile-time assertions verify both satisfy them). The connector provider is supplied later by the REST layer (T8).

### Tests (TDD, `-tags server -race`)
Each ordering branch, near-expiry refresh (token-exchange **and** connect-flow), unconnected→connect-URL error, unauthenticated reject, store-disabled, no-broker-config, policy-denied, no-static-fallback on exchange failure, and single-flight coalescing under `-race`.

### Verification
- `go test -tags server ./internal/serveredition/broker/ -race` ✅
- `go build -tags server ./cmd/mcpproxy` ✅ + `go build ./cmd/mcpproxy` (personal unaffected) ✅
[...]

PR #690, opened 2026-06-15
test(broker): direct cross-user isolation test for CredentialResolver (MCP-2578)

Dumbris:
## Summary
Backlog follow-up to **MCP-1039 / #688** (merged). #688 verified per-user isolation only **structurally** — every lookup is `store.Get(userID, serverKey)`-keyed with no shared fallback — but there was **no explicit test** seeding user B's credential and asserting user A's `Resolve` never returns it.

This PR adds `TestResolve_CrossUserIsolation_NeverReturnsAnotherUsersCredential` in `internal/serveredition/broker` (build tag `server`):

- Seeds a **valid, non-expiring** credential for **userB + serverKey**.
- Calls `Resolve` as **userA** for the same serverKey.
- Asserts userA gets the **fail-closed path** and **never** userB's token:
  - **token_exchange** mode → propagated exchange error, `got == nil`, and userA goes through its *own* acquisition (1 `Exchange` call, proving no cache short-circuit to userB).
  - **oauth_connect** mode → `*NotConnectedError` with userA's connect URL, no refresh against userB's cached credential, error never contains userB's token.

## Why it's a real regression guard
The seeded userB credential is valid and not near expiry, so a regression dropping per-user keying (a shared/static fallback) would make `acquire()` return it directly from cache. Verified by temporarily simulating a shared fallback — both sub-tests fail and userA literally receives `userB-secret-token-MUST-NOT-LEAK`.

## Verification
- `go test -tags server ./internal/serveredition/broker/... -race` → green
- `golangci-lint` (CI v2 config, `--build-tags server`) → 0 issues
- Test-only change; no production code touched.

Related #688

### DIFF
--- a/internal/serveredition/broker/credential_resolver_test.go
+++ b/internal/serveredition/broker/credential_resolver_test.go
@@ -483,3 +483,97 @@ func TestResolve_ConnectFlow_RefreshFails_ReturnsReconnectError(t *testing.T) {
 		t.Fatalf("expected the connect URL to be built once, got %d", c)
 	}
 }
+
+// TestResolve_CrossUserIsolation_NeverReturnsAnotherUsersCredential is the
+// direct cross-user isolation guard (MCP-2578, backlog follow-up to MCP-1039 /
+// #688). #688 verified isolation only structurally (every lookup is
+// store.Get(userID, serverKey)-keyed with no shared fallback); this asserts the
+// behaviour end-to-end: seed user B's credential, Resolve as user A for the
+// SAME serverKey, and prove user A gets the fail-closed path and NEVER user B's
+// token.
+//
+// The seeded user B credential is deliberately VALID and not near expiry, so a
+// regression that dropped the per-user keying (a shared/static fallback) would
+// make acquire() return it directly from cache (FR-014). With correct keying,
+// store.Get(userA, key) misses and user A falls through to its own — absent —
+// acquisition path, which fails closed.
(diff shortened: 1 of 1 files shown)


Dumbris:
**Code Review — CEO Agent (MCP-2582)**

Test-only PR. Verified locally: `go test -tags server ./internal/serveredition/broker/... -run TestResolve_CrossUserIsolation -race` → **ok**

**Assessment: ACCEPT**

Test design is correct and the regression guard is well-constructed:
- Valid long-lived userB seed ensures a shared-cache regression would leak the credential (not a vacuous pass)
- `token_exchange` branch: 1 Exchange call proves userA traversed its own acquisition path, not userB's cache
- `oauth_connect` branch: 0 refresh calls + error text check closes the leak surface
- Seed-intact assertion rules out false-positives from universal cache misses

No production code touched. Ready to merge.

Dumbris:
**CEO Code Review — LGTM** (CodexReviewer exhausted credits; reviewed as MCP-2587 backstop)

**Two subtests, both PASS** (`-tags server`):
- `token_exchange mode falls closed, not to user B's cache` ✅
- `oauth_connect mode falls closed to NotConnectedError, not user B's cache` ✅

**Review notes:**
- Adversarially structured: user B's credential is valid + long-lived, so a shared-cache regression would return it directly and skip Exchange entirely — the `ex.calls == 1` assertion catches exactly that regression.
- Both acquisition modes covered (token_exchange + oauth_connect).
- User B's credential verified intact post-run — proves the miss is per-user isolation, not data absence.
- No production code changes; 94-line pure test addition.

Ready to merge.

PR #698, opened 2026-06-17
fix(frontend): security scan UI stuck on 'Scanning…' though backend job completed (MCP-2740)

Dumbris:
## Summary
Fixes the Server Detail → Security tab staying stuck on **"Scanning… N/N complete"** with the **Report button disabled indefinitely**, even though the backend scan job is already `completed` (observed: backend finished in ~1.5s, UI stuck ~25 min).

**Root cause** (`frontend/src/views/ServerDetail.vue`): `scanLoading` was only ever cleared inside a live poll tick whose job id matched `activeScanJobId`. A sub-2s scan can finalize — and be replaced by a Pass-2 job, or have its id drift — before the first 2s poll tick, so the `jobId !== activeScanJobId && scan_pass !== 2` branch `return`ed without ever clearing the spinner. There was no idempotent reconciliation against the authoritative backend status.

## Fix
Derive terminal state from the **authoritative backend status on every status fetch**, independent of `activeScanJobId` / `scan_pass` bookkeeping:

- **New pure reducer** `frontend/src/utils/scanState.ts` — `isTerminalScanStatus()` + `decideScanReconcile()` (fully unit-tested, 10 cases).
- **Poll tick & initial `loadScanReport`** both finalize on any terminal status (`completed`/`complete`/`failed`/`error`/`cancelled`) or a newer running Pass-2 job, regardless of job-id mismatch.
- **`loadScanReport`** clears a stale `scanLoading` when the latest job is already terminal (a scan that finished while the component was unmounted).
- **Render-independent safety-net watcher** so the spinner can never outlive a terminal status.
- **`data-test` hooks** added: `scan-button`, `scan-progress`, `scan-report-link`.

Behavior preserved: the two-pass scan flow, the "Scan Complete" toast (now guarded so it only fires when a scan was actually in flight), and the Pass-2 handling.

## Verification
- `vitest`: full suite **197/197 pass**, incl. new `tests/unit/scan-state-reconcile.spec.ts`.
- `vue-tsc --noEmit`: clean. `make build`: clean (frontend embedded).
- **Playwright sweep against the built (embedded) UI** reproducing the exact race (start → `job-A`; completing `/scan/status` poll → `job-B`, `completed`): spinner clears, button returns to "Scan Now", Report link enables. Plus fresh-mount-on-completed scenario. Report + screenshots + spec under `specs/mcp-2740-scan-ui-stuck/verification/`.

Scope note: the 3/5-failed-scanner findings in the original report (ElevenLabs `uvx` spawn, graceful-skip for missing `tools.json`, ramparts arm64 glibc) are **explicitly out of scope** for this issue and not addressed here.

Related MCP-2740


Dumbris:
## 🤖 Codex review (model gpt-5.5, grounded in `fix/mcp-2740-scan-ui-stuck` vs `origin/main`)

**Verdict: REQUEST_CHANGES** — one real correctness gap (P2); the core reconciliation approach is otherwise sound.

### [P2] Pass-2 reload can re-enable the "Scanning…" state — `frontend/src/views/ServerDetail.vue:2821`
When `finalizeScan` is triggered by a *newer Pass-2 job* (`scan_pass === 2`), it clears `activeScanJobId` and then calls `loadScanReport(true)`. While Pass 2 is active, `GET /scan/status` returns that Pass-2 job as `running`, so `loadScanReport`'s `decideScanReconcile` takes the `resumePolling` branch and sets `scanLoading` back to `true` — hiding/disabling the completed Pass-1 report until Pass 2 finishes. This preserves a variant of the stuck-spinner behavior the PR is fixing.

**Fix options:** suppress status reconciliation for this post-finalize report reload, or don't resume polling for `scan_pass === 2` jobs in that path (e.g. pass a flag into `loadScanReport` to skip the resume-polling branch when reloading after a Pass-1 finalization).

_Reviewed via `codex exec review --base origin/main` in a checkout of the PR branch._

Dumbris:
**Fixed P2 (Codex finding):** Pass-2 reload re-enabling `scanLoading`

Root cause confirmed: `finalizeScan` called `loadScanReport(true)` which fetched scan status, got the Pass-2 job as `running`, and `decideScanReconcile` correctly returned `resumePolling: true` — but without a guard the caller re-set `scanLoading = true` and hid the Pass-1 report.

Fix (commit `90ff3e7e`):
- Added `skipPolling = false` parameter to `loadScanReport`
- `finalizeScan` now calls `loadScanReport(true, true)` — fetches the report data without starting a new poll cycle
- Same for the safety-net watcher (same post-finalize pattern)
- Unit test added documenting the post-finalize cleared-state invariant: `decideScanReconcile` with `{scanLoading:false, activeScanJobId:null}` + running Pass-2 returns `resumePolling:true` — callers after `clearScanRunState` must suppress it

198/198 tests pass, `make build` clean.

Dumbris:
## 🤖 Codex re-review — ACCEPT ✅ (1 non-blocking P3)

Re-reviewed after the Pass-2 fix (`skip polling on post-finalize reload`).

> The main stuck-spinner fix is sound.

No P0–P2 findings. One **[P3]** (cosmetic, non-blocking) — `frontend/src/utils/scanState.ts:68`: `cancelled`/`canceled` statuses currently take the success path in `finalizeScan`, so a scan cancelled from another tab/during a cancel-race can flash "Scan Complete" and reload a possibly-stale report. Filed as a follow-up; not blocking this merge.

_Reviewed via `codex exec review --base origin/main` on `90ff3e7e`._

PR #701, opened 2026-06-17
feat(shellwrap): hydrate login-shell env once at startup (macOS GUI/launchd) — MCP-2751

Dumbris:
## Summary

Fixes the **root-cause class** behind the recent docker/PATH point-fixes (#697 resolve-absolute docker path, #699 DOCKER_* capture): when mcpproxy is launched from a **macOS GUI/launchd context** (Launchpad/Dock, the `SMAppService` login item, or the tray spawning the core), the Go core inherits a **launchd-minimal environment**. It never sources the user's login shell, so it lacks Homebrew/Docker PATH entries, `docker`/`uvx`/`npx` on PATH, and exported vars like `DOCKER_HOST`/`DOCKER_CONTEXT`. Launched from a terminal it's healthy.

This adds a **one-time, allow-listed login-shell environment hydration at startup** so every spawn path inherits a correct environment without per-spawn `$SHELL -lc` wrapping — the durable consolidation that supersedes per-var capture.

## What changed

- **`internal/shellwrap/hydrate.go`** (new): `captureLoginShellEnv()` sources `$SHELL -l -c 'env -0'` **once** (NUL-delimited so multiline/`=`-containing values survive; marker-bracketed to strip rc-file banner noise; 5s timeout; `sync.Once`) and caches the parsed env. `LoginShellPATH` now reads PATH from this **shared capture — one shell fork, not two**.
- **`HydrateFromLoginShell(logger)`**: gates on macOS **and** launchd-minimal PATH (no-op on terminal launches and non-macOS), merges `PATH` login-first, and applies a curated allow-list set-if-empty via `os.Setenv`:
  - `DOCKER_HOST`, `DOCKER_CONTEXT`, `DOCKER_CONFIG`, `DOCKER_CERT_PATH`, `DOCKER_TLS_VERIFY`
  - `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` (+ lowercase)
  - `NVM_DIR`, `ASDF_DIR`, `PYENV_ROOT`, `VOLTA_HOME`, `HOMEBREW_PREFIX`, `COLIMA_HOME`
- **`internal/secureenv/manager.go`**: extend the default allow-list with the curated keys so the hydrated vars survive the filter into upstream stdio/docker spawns. Scanner `MinimalEnv` stays narrow (credential-free).
- **`cmd/mcpproxy/main.go`**: call `HydrateFromLoginShell` after logger setup, before any manager reads `os.Environ()` — one `os.Setenv` hydration propagates to all spawn paths.
- **docs**: `MCPX_STDIO_SPAWN_ENOENT.md` updated to describe the automatic startup hydration.

## Security / privacy

- **Allow-list, never wholesale `env` import** — secrets (`AWS_*`, `GITHUB_TOKEN`, `ANTHROPIC_API_KEY`, …) are never pulled into the daemon or subprocesses (unit-tested).
- **Set-if-empty** — never clobbers an operator-set value. `HOME`/`USER`/`SHELL` never touched.
- **Values never logged** — key names + lengths only (privacy unit test).

## Relationship to #697 / #699

- **#697 kept** — the well-known-path docker probe is a complementary fallback.
- **#699 env part superseded** — hydration now places curated `DOCKER_*` into the process env; per-spawn capture is redundant. (Absolute-path resolve + negative-cache fix remain.)

## Tests / verification

- `internal/shellwrap`: gate no-op (non-darwin + comprehensive PATH), curated-set hydration, set-if-empty never-clobber, never-touch HOME/USER/SHELL, secret-exclusion, privacy (no values logged), capture-failure no-op. Gate seam (`hydrationGOOS`) exercises the logic on Linux CI.
- `internal/secureenv`: curated DOCKER_*/proxy vars pass the allow-list while secrets stay filtered; existing `TestLaunchdMinimalPath_*` gate guarantees preserved.
- `go test ./internal/shellwrap/... ./internal/secureenv/... ./internal/upstream/core/... -race` — green.
- `golangci-lint v2` (CI config) — 0 issues. Both editions build.
- **Real-binary check**: launched with a simulated launchd-minimal `PATH` → hydration fires and enriches PATH; launched with a comprehensive `PATH` → no-op (gate confirmed).

Related #699

### DIFF
--- a/cmd/mcpproxy/main.go
+++ b/cmd/mcpproxy/main.go
@@ -47,6 +47,7 @@ import (
 	"github.com/smart-mcp-proxy/mcpproxy-go/internal/logs"
 	"github.com/smart-mcp-proxy/mcpproxy-go/internal/registries"
 	"github.com/smart-mcp-proxy/mcpproxy-go/internal/server"
+	"github.com/smart-mcp-proxy/mcpproxy-go/internal/shellwrap"
 	"github.com/smart-mcp-proxy/mcpproxy-go/internal/storage"
 	"github.com/smart-mcp-proxy/mcpproxy-go/internal/telemetry"
 	_ "github.com/smart-mcp-proxy/mcpproxy-go/oas" // Import generated swagger docs
@@ -498,6 +499,17 @@ func runServer(cmd *cobra.Command, _ []string) error {
 		zap.String("edition", Edition),
 		zap.String("log_level", cmdLogLevel),
 		zap.Bool("log_to_file", cmdLogToFile))
+
+	// MCP-2751: when launched from a macOS GUI/launchd context (Launchpad, the
+	// SMAppService login item, or the tray spawning the core), the process
+	// inherits a launchd-minimal environment and never sources the user's
+	// login shell — so it lacks Homebrew/Docker PATH entries and exported vars
+	// like DOCKER_HOST. Hydrate a curated allow-list (PATH + DOCKER_*/proxy/
+	// tool-home) once, before any manager reads os.Environ(), so every spawn
+	// path (docker, stdio servers, uvx/npx, ResolveDockerPath,
+	// secureenv.BuildSecureEnvironment) inherits a correct environment with no
+	// call-site changes. No-op on terminal launches and non-macOS.
+	shellwrap.HydrateFromLoginShell(logger)
 
 	// Pass edition and version to internal packages
 	httpapi.SetEdition(Edition)
--- a/docs/errors/MCPX_STDIO_SPAWN_ENOENT.md
+++ b/docs/errors/MCPX_STDIO_SPAWN_ENOENT.md
@@ -55,8 +55,17 @@ sudo apt install nodejs npm  # Debian/Ubuntu
 
 ### 2. Make the GUI inherit the shell PATH (macOS)
 
-The macOS tray launches mcpproxy with the user's login `PATH`, which is shorter
-than your interactive shell `PATH`. Either:
+When launched from a macOS GUI/launchd context (Launchpad, the login item, or
+the tray spawning the core), mcpproxy inherits a launchd-minimal environment
+rather than your interactive shell `PATH`. Since the daemon now **hydrates the
+login-shell environment once at startup** — sourcing your login shell to merge
+in `PATH` (and curated `DOCKER_*`/proxy/tool-home vars) — `uvx`/`npx` installed
+in `/opt/homebrew/bin`, `~/.local/bin`, or a version-manager shim directory are
+normally found automatically. (Hydration is a no-op when mcpproxy is started
+from a terminal whose `PATH` is already comprehensive.)
+
+If the tool still isn't found — e.g. it lives in a directory your login shell
+doesn't export, or your rc files don't run non-interactively — either:
 
 - Move the binary to a system directory: `sudo ln -s "$(which uvx)" /usr/local/bin/uvx`, or
 - Set an absolute path in the upstream config: `"command": "/Users/you/.local/bin/uvx"`.
--- a/internal/secureenv/launchd_path_test.go
+++ b/internal/secureenv/launchd_path_test.go
@@ -215,12 +215,12 @@ func TestLaunchdMinimalPath_AlreadyComprehensive(t *testing.T) {
 		"comprehensive PATH must be returned unchanged — terminal-launched processes should not be polluted by login-shell capture")
 }
 
-// TestBuildSecureEnvironment_AllowsHydratedDockerAndProxyVars verifies the
-// allow-list extension for MCP-2751: once shellwrap.HydrateFromLoginShell has
-// placed curated DOCKER_*/proxy vars into the process env, the default
-// allow-list must pass them through to spawned upstreams — while genuine
-// secrets remain filtered out.
-func TestBuildSecureEnvironment_AllowsHydratedDockerAndProxyVars(t *testing.T) {
+// TestBuildSecureEnvironment_AllowsHydratedDockerVars verifies the allow-list
+// extension for MCP-2751: once shellwrap.HydrateFromLoginShell has placed
+// curated DOCKER_*/tool-home vars into the process env, the default allow-list
+// must pass them through to spawned upstreams — while secrets remain filtered.
+// Proxy vars are intentionally excluded from the allow-list (separate follow-up).
+func TestBuildSecureEnvironment_AllowsHydratedDockerVars(t *testing.T) {
 	if runtime.GOOS == "windows" {
 		t.Skip("uses unix PATH semantics")
 	}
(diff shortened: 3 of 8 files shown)


Dumbris:
## 🤖 Codex review — REQUEST_CHANGES (1×P2)

### [P2] Respect explicitly-empty operator env vars — `internal/shellwrap/hydrate.go:234-236`
The set-if-empty guard treats an explicitly-set empty var the same as unset, so an operator who sets `HTTPS_PROXY=` or `DOCKER_HOST=` to *disable* an inherited login-shell value will have it overwritten from the login shell — violating the "never clobber operator-set values" contract. Use `os.LookupEnv` to distinguish unset from intentionally-empty before hydrating.

_Reviewed via `codex exec review --base origin/main`._

Dumbris:
## Addressed — Codex P2 (set-if-empty vs intentionally-empty)

Valid catch, fixed in e42f22fd. The curated-key guard used `os.Getenv(key) != ""`, which can't distinguish *unset* from *explicitly set-empty* — so an operator who sets `HTTPS_PROXY=` / `DOCKER_HOST=` to disable an inherited value would have it clobbered from the login shell.

**Fix:** switched the guard to `os.LookupEnv(key)` — any **present** key (even empty) is treated as operator intent and preserved. launchd never sets these to empty, so present-but-empty is always deliberate.

**Test:** added `TestHydrate_PreservesIntentionallyEmptyVar` — sets `HTTPS_PROXY=""`, has the login shell export a non-empty proxy, and asserts the var stays present-and-empty (not overwritten) and is absent from the diagnostic snapshot. (Under the old guard this test fails.)

`go test ./internal/shellwrap/... -race` green; golangci-lint v2 (CI config) 0 issues.

Dumbris:
## 🤖 Codex re-review (round 2) — REQUEST_CHANGES (1×P2)

### [P2] Treat proxy spellings as aliases before hydrating — `internal/shellwrap/hydrate.go:241-242`
When the daemon already has a proxy var in the *opposite case* (e.g. `https_proxy=` to disable, or `https_proxy=http://operator`), the exact-key check for `HTTPS_PROXY` misses it and imports the login-shell variant. HTTP clients honor either spelling, so a downstream MCP server can end up using the login-shell proxy despite an operator-set value. Check **all case variants** of the logical proxy var before hydrating either spelling.

_Reviewed via `codex exec review --base origin/main` on `e42f22fd`._

Dumbris:
## Addressed — Codex round-2 P2 (proxy case-aliasing)

Correct again, fixed in 3983b7e3. Since HTTP clients honor either spelling, the two case variants of each proxy var are one logical var. The per-key `LookupEnv` still missed the opposite-case operator value.

**Fix:** split the allow-list into single-spelling keys (DOCKER_*, tool-home) and an alias-aware proxy trio. For each logical proxy var I now check **both** `UPPER` and `lower` via `os.LookupEnv` before hydrating; if **either** is present (including intentionally empty), I skip hydrating **both** spellings. DOCKER_*/tool-home keep the single-spelling set-if-unset path.

**Test:** added `TestHydrate_ProxyCaseAliasingPreservesOperatorIntent` — daemon has `https_proxy=''` (lowercase empty), login shell exports `HTTPS_PROXY=http://x` → neither spelling imported, operator's empty lowercase preserved. (Fails under the previous per-key guard.)

`go test ./internal/shellwrap/... ./internal/secureenv/... -race` green; golangci-lint v2 (CI config) 0 issues; both editions build.

Dumbris:
## 🤖 Codex re-review (round 3) — REQUEST_CHANGES (2×P2)

### [P2] Don't gate hydration on PATH alone — `internal/shellwrap/hydrate.go:212-217`
A GUI launch can have PATH pre-seeded (e.g. `launchctl config user path` includes `/opt/homebrew/bin`) yet still lack `DOCKER_HOST`/tool-home vars; this early-return then skips hydration and leaves the non-PATH part unresolved. Trigger hydration when curated vars are missing, not only when PATH looks minimal.

### [P2] Credential leak — proxy vars in the default allow-list — `internal/secureenv/manager.go:95-98`
Adding `HTTP_PROXY`/`HTTPS_PROXY` to the default allow-list forwards credential-bearing proxy URLs (`http://user:pass@proxy`) to **every** stdio upstream on **all** platforms.

**Resolution (scope reduction): drop proxy vars from this PR entirely** — remove `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` from both hydration and the secureenv allow-list. Proxy forwarding is out of scope for MCP-2751 (it's the source of the round-2 alias churn and this credential leak) and will be a separate opt-in feature. Keep PATH + `DOCKER_*` + tool-homes, and fix the launch gate above.

_Reviewed via `codex exec review --base origin/main` on `3983b7e3`._

Dumbris:
**Round-3 fix pushed (`5f0f8617`)**

Two changes as directed:

**1. Proxy vars dropped entirely**
- Removed `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` and the whole `proxyVarPairs` alias logic from hydration
- Removed them from `secureenv.DefaultEnvConfig` allow-list (`manager.go`)
- Removed tests: `PreservesIntentionallyEmptyVar`, `ProxyCaseAliasingPreservesOperatorIntent`
- Added negative assertion in `MinimalPathHydratesCuratedSet` that proxy vars are NOT imported

**2. Launch gate fixed for pre-seeded PATH**
- Gate now: `pathIsMinimal || anyDockerVarMissing()`
- PATH-merge stays gated on `pathIsMinimal` only (comprehensive PATH is not widened)
- Curated-var capture runs whenever either condition triggers
- `anyDockerVarMissing()` checks only `DOCKER_\*` prefixed entries in `curatedSingleKeys`
- Renamed `GateNoOpOnComprehensivePath` → `GateNoOpOnComprehensivePathAndAllDockerPresent` (now pre-sets all 5 DOCKER_* vars as the true no-op condition)
- Added `ComprehensivePathStillHydratesDockerIfMissing` — comprehensive PATH + missing DOCKER_HOST → hydrates docker vars but leaves PATH unmodified

**Scope: PATH + DOCKER_* + tool-homes only.** Local: `go test -race ./internal/shellwrap/... ./internal/secureenv/...` green, golangci-lint v2 0 issues.

Dumbris:
## 🤖 Codex review — ACCEPT ✅

> The changed code is internally consistent, covered by targeted tests, and I did not identify any discrete regression that would break existing behavior.

No P0–P2 findings after the scope reduction (proxy vars dropped → no credential-leak surface; launch gate no longer relies on PATH alone). Login-shell hydration of PATH + `DOCKER_*` + tool-homes for macOS GUI/launchd launches. Proxy forwarding split to a separate opt-in follow-up (MCP-2769).

_Reviewed via `codex exec review --base origin/main` on `5f0f8617`._

PR #703, opened 2026-06-17
fix(docker): direct-exec resolved docker binary for isolated spawn (MCP-2753)

Dumbris:
## Summary

Systematic follow-up deferred from #699 / MCP-2744 (board kept #699 minimal). Migrates the docker-isolation spawn from a `$SHELL -l -c "<docker> run …"` login-shell wrap to **direct exec of the resolved absolute docker binary**, so the spawn bypasses `PATH` entirely.

PR #697 resolved the docker CLI to an absolute path but still routed the spawn through the login shell, where `$PATH` is re-derived from rc files and can drop the Docker Desktop bundle dir — so docker-isolated servers still hit `command not found: docker` despite a good `docker_path`. Exec'ing the absolute path directly (mirroring the management path's `newDockerCmd`) fixes that.

## What changed
- `setupDockerIsolation` now returns `(cmd, args, shellWrapped bool)`. On resolution to a **verified absolute executable** it returns the resolved binary + raw docker argv with `shellWrapped=false`.
- New `isDirectExecutable` guard (**Codex round-3 P2 #1**): `filepath.IsAbs` + `os.Stat` + exec-bit. `ResolveDockerPath`'s last resort runs `command -v docker` in the login shell, which can emit a shell function/alias/bare name — such a value is rejected and shell-wrapped instead of failing a direct exec.
- New args-based `insertCidfileIntoDockerArgs` (inserts `--cidfile` after the `run` token; platform-agnostic, sidesteps the `-c` vs `/c` shell quirk). Both call sites (`connection_stdio.go`, `connection_launcher.go`) branch on `shellWrapped` to pick the matching cidfile helper.
- Login-shell wrap of bare `docker` retained as the resolution-failure / non-executable fallback.
- `docs/docker-isolation.md` updated to describe direct-exec behavior.

## Scope #2 (DOCKER_* env hydration) — already shipped
The issue's second scope item (login-shell DOCKER_* hydration) is **satisfied by the merged MCP-2751 / #701**, which hydrates `DOCKER_*` into `os.Environ()` at startup and carries them through `BuildSecureEnvironment`'s default allow-list. So the direct-exec'd docker process still reaches Colima/rootless/remote daemons with no per-spawn shell capture. The reverted prior-art commits `b54d0b07` (per-spawn allowlist) and `c9eac4fc` (per-spawn capture) are obsolete and were **not** recovered — #701 explicitly supersedes them.

> Note on Codex P2 #2 (scope DOCKER_* to docker servers only): #701 deliberately placed the curated `DOCKER_*` keys in the secureenv **default** allow-list (treated as connection pointers, not secrets) and was reviewed/merged on that basis. Re-scoping to docker-only would re-litigate a shipped decision and break #701's LaunchAgent recovery, so it is intentionally out of scope here. Flagging for reviewer visibility.

## Verification
- `go test ./internal/upstream/core/... ./internal/shellwrap/... ./internal/secureenv/... -race` → all green
- `golangci-lint` (CI v2 config) on the three packages → 0 issues
- `go build ./cmd/mcpproxy`, `go vet`, `gofmt` clean
- New tests: direct-exec of verified abs executable, shell-wrap fallback for non-absolute and non-existent resolved values, `insertCidfileIntoDockerArgs` unit table.

Related #699
Related #696

### DIFF
--- a/docs/docker-isolation.md
+++ b/docs/docker-isolation.md
@@ -251,11 +251,19 @@ docker stats
   it is **not** on a standard `PATH` dir unless you ran the optional,
   admin-gated "install CLI tools" step. When mcpproxy is launched from a
   LaunchAgent / tray, the captured login-shell `PATH` may omit this directory.
-- mcpproxy resolves the `docker` binary to its **absolute path** before
-  spawning an isolated server (and also adds the bundle bin dir to the enhanced
-  spawn `PATH`), so isolation works even without the CLI-tools step. If you
-  still see this error, confirm the binary exists at the bundle path above, or
-  run Docker Desktop's "install CLI tools".
+- mcpproxy resolves the `docker` binary to its **absolute path** and then
+  **exec's it directly** (no login-shell wrap) when spawning an isolated server,
+  so the spawn bypasses `PATH` entirely and works even without the CLI-tools
+  step. (The enhanced spawn `PATH` still includes the bundle bin dir as a
+  belt-and-suspenders measure.) Earlier builds resolved the absolute path but
+  still routed the spawn through `$SHELL -l -c "<docker> run …"`, where the
+  login shell re-derived `PATH` from rc files and could drop the bundle dir —
+  so the error persisted; direct exec fixes that. Direct exec is used only when
+  the resolved value is a verified absolute executable; a non-absolute result
+  (e.g. a shell function/alias from `command -v docker`) falls back to the
+  login-shell wrap of bare `docker`. If you still see this error, confirm the
+  binary exists at the bundle path above, or run Docker Desktop's "install CLI
+  tools".
 - `upstream_servers list` reports `docker_status.docker_path` (the resolved
   binary) and reports `docker_status.available` / per-server `docker_available`
   as `true` **only** when the CLI is actually resolvable *and* `docker info`
--- a/docs/docker-isolation.md
+++ b/docs/docker-isolation.md
@@ -259,11 +259,16 @@ docker stats
   still routed the spawn through `$SHELL -l -c "<docker> run …"`, where the
   login shell re-derived `PATH` from rc files and could drop the bundle dir —
   so the error persisted; direct exec fixes that. Direct exec is used only when
-  the resolved value is a verified absolute executable; a non-absolute result
-  (e.g. a shell function/alias from `command -v docker`) falls back to the
-  login-shell wrap of bare `docker`. If you still see this error, confirm the
-  binary exists at the bundle path above, or run Docker Desktop's "install CLI
-  tools".
+  (a) the resolved value is a verified absolute executable and (b) the docker
+  daemon-config env is guaranteed without the login shell — on macOS via the
+  startup login-shell hydration, or on any platform when `DOCKER_HOST` /
+  `DOCKER_CONTEXT` are already exported into mcpproxy's environment. A
+  non-absolute result (e.g. a shell function/alias from `command -v docker`), or
+  a rootless/remote daemon on Linux whose `DOCKER_HOST` lives only in the
+  login-shell rc, falls back to the `$SHELL -l` wrap (still using the resolved
+  absolute path when one was found) so `docker run` keeps inheriting the daemon
+  config. If you still see this error, confirm the binary exists at the bundle
+  path above, or run Docker Desktop's "install CLI tools".
 - `upstream_servers list` reports `docker_status.docker_path` (the resolved
   binary) and reports `docker_status.available` / per-server `docker_available`
   as `true` **only** when the CLI is actually resolvable *and* `docker info`
--- a/internal/upstream/core/connection_docker.go
+++ b/internal/upstream/core/connection_docker.go
@@ -2,6 +2,9 @@ package core
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/smart-mcp-proxy/mcpproxy-go/internal/shellwrap"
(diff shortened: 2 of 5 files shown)


Dumbris:
## 🤖 Codex review — REQUEST_CHANGES (1×P2)

### [P2] Keep login-shell Docker vars before direct-exec on non-Darwin — `internal/upstream/core/connection_docker.go:96-97`
On **non-Darwin** launches where `docker` resolves but daemon selection is exported only by the login shell (e.g. rootless Docker with `DOCKER_HOST` in `.profile`), this direct-exec branch bypasses the old `wrapWithUserShell` path. `HydrateFromLoginShell` is **Darwin-only** and `BuildSecureEnvironment` only forwards `DOCKER_*` already present in the process env — so `docker run` can lose the daemon config and fail even though the shell-wrapped spawn worked.

**Fix options:** gate the direct-exec to cases where env is guaranteed (Darwin where hydration ran, **or** `DOCKER_HOST`/`DOCKER_CONTEXT` already present in `os.Environ()`), and keep the `wrapWithUserShell` fallback otherwise; or extend login-shell `DOCKER_*` capture to non-Darwin. Add a non-Darwin test: `DOCKER_HOST` only via login shell → spawn still receives it (or stays shell-wrapped).

_Reviewed via `codex exec review --base origin/main` on `7fe593a2`._

Dumbris:
## Addressed Codex P2 — non-Darwin rootless daemon-env regression (`04dbe4b0`)

Confirmed the concern is real and verified the premises against `origin/main`:
- `HydrateFromLoginShell` is **Darwin-only** (`hydrationGOOS != osDarwin → no-op`), gate `pathIsMinimal || anyDockerVarMissing()`.
- `BuildSecureEnvironment` only forwards `DOCKER_*` already in `os.Environ()`.
- So on Linux, a daemon whose `DOCKER_HOST`/`DOCKER_CONTEXT` lives only in `.profile` is lost when direct-exec drops the `$SHELL -l` wrap.

### Fix (your primary recommendation — gate, don't extend capture)
Added a second precondition `dockerDaemonEnvGuaranteed()` alongside the abs-executable check. Direct-exec now requires **both**:
1. `isDirectExecutable(resolved)` — verified absolute executable (P2 #1), and
2. daemon env guaranteed without the login shell — `GOOS == darwin` (startup hydration captured `DOCKER_*` into `os.Environ()`) **OR** `DOCKER_HOST`/`DOCKER_CONTEXT` already in `os.Environ()` on any platform.

Otherwise we keep the `$SHELL -l` wrap so `docker run` inherits rc-file `DOCKER_*`. The fallback now wraps the **resolved absolute path** (still sidesteps the login shell's PATH re-derivation), dropping to bare `docker` only when resolution failed.

I went with gating rather than extending the login-shell `DOCKER_*` capture to Linux — that keeps #701's deliberate "Darwin-only hydration" scope intact and avoids re-introducing a per-spawn fork on Linux.

### Tests (your requested non-Darwin case + more)
- `TestSetupDockerIsolationShellWrapsWhenDaemonEnvMissingNonDarwin` — GOOS=linux, `DOCKER_HOST`/`DOCKER_CONTEXT` unset → stays **shell-wrapped** with the absolute path (so the login shell recovers the daemon config).
- `TestSetupDockerIsolationDirectExecsWhenDockerHostInEnv` — GOOS=linux, `DOCKER_HOST` set → direct-execs.
- `TestDockerDaemonEnvGuaranteed` — table test of the gate. `dockerDaemonEnvGOOS` is a package var so the Darwin branch is exercised on the Linux CI host.

`go test ./internal/upstream/core/... ./internal/shellwrap/... ./internal/secureenv/... -race` green; golangci-lint v2 clean; docs updated.

Dumbris:
## ✅ Review (Claude Code, Codex rate-limited) — ACCEPT

Rigorous substitute review while Codex is rate-limited (per maintainer direction).

**Previously-flagged P2 (non-Darwin rootless regression) — CONFIRMED RESOLVED.** Direct-exec is gated in `setupDockerIsolation` (`connection_docker.go:110-137`): direct-exec only when `resErr==nil && isDirectExecutable(resolved) && dockerDaemonEnvGuaranteed()`. `dockerDaemonEnvGuaranteed()` (`:178-183`) is true on Darwin (startup hydration MCP-2751 forwards `DOCKER_*`) and on non-Darwin only when `DOCKER_HOST`/`DOCKER_CONTEXT` are already in `os.Environ()`; otherwise it falls back to `wrapWithUserShell` (wrapping the **resolved absolute path**, not bare `docker`), preserving the Linux/rootless `.profile`-`DOCKER_HOST` behavior #699 protected. `HydrateFromLoginShell` confirmed Darwin-only, so the gate's asymmetry is justified.

**Other points ✓** — `isDirectExecutable` requires absolute path + existing non-dir file + exec bit (non-absolute `command -v` results fall back to shell). macOS Docker Desktop case still direct-execs the bundle path (no regression). Cidfile: args-based insert on direct-exec, `-c`/`/c`-tolerant shell insert on fallback; both call sites (`connection_stdio.go:153-159`, `connection_launcher.go:252-258`) branch on `dockerShellWrapped` correctly. Container-tracking unchanged.

**Tests ✓** — new guards `TestSetupDockerIsolationShellWrapsWhenDaemonEnvMissingNonDarwin`, `...DirectExecsWhenDockerHostInEnv`, `TestDockerDaemonEnvGuaranteed` (4 subtests), plus existing direct-exec/non-absolute/non-executable/cidfile/Windows-`/c` guards. `go test ./internal/upstream/core/...` green; `go build ./...` exit 0.

**VERDICT: ACCEPT**

PR #704, opened 2026-06-17
feat(secureenv): opt-in proxy env forwarding to upstreams (MCP-2769)

Dumbris:
## Summary

Split out of MCP-2751 (login-shell hydration). Codex flagged that adding `HTTP_PROXY`/`HTTPS_PROXY` to the secureenv default allow-list would forward **credential-bearing** proxy URLs (`http://user:pass@proxy`) to **every** stdio upstream on **all** platforms. This PR keeps proxy vars out of the default allow-list and makes forwarding an explicit, credential-redacting opt-in.

## Behavior

- New global config flag **`forward_proxy_env`** (default `false`). Current behavior is unchanged when unset.
- When `true`, `secureenv.BuildSecureEnvironment` forwards the ambient proxy variables to spawned stdio upstreams:
  - Vars: `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, `ALL_PROXY`, `FTP_PROXY` — both upper- and lower-case spellings (case-alias aware).
  - **Userinfo (credentials) is redacted** from each value (`http://user:pass@proxy:8080` → `http://proxy:8080`); the proxy host/port is preserved.
  - An explicitly configured proxy value (`custom_vars` or a server's `env`) wins and suppresses forwarding of the ambient value, including the other-cased alias.
  - Forwarding is independent of `inherit_system_safe` (it is its own explicit opt-in).
- Proxy vars remain **excluded from the default allow-list** and from login-shell hydration (unchanged from #701).

## Files

- `internal/config/config.go` — `forward_proxy_env` flag.
- `internal/secureenv/manager.go` — forwarding + `redactProxyCredentials`.
- `internal/upstream/core/client.go` — thread flag into per-server stdio env.
- `docs/configuration.md` — opt-in docs + never-by-default rationale.
- `oas/*` — regenerated (config is a documented model).

## Tests

- `redactProxyCredentials` table (user:pass, user-only, socks, host-list, unparseable).
- Default-off **leak guard** (proxy in process env is NOT forwarded).
- Opt-in forwards with credentials redacted; `NO_PROXY` passes unchanged.
- Explicit value wins over the ambient alias.
- End-to-end wiring test through `managed.NewClient`.

## Verification

```
go test ./internal/secureenv/... ./internal/config/... ./internal/upstream/... -race   # ok
golangci-lint v2.5.0 run --config .github/.golangci.yml <changed pkgs>                 # 0 issues
go build ./... && go vet <changed pkgs>                                                # ok
```

Related #701


Dumbris:
## 🤖 Codex review — REQUEST_CHANGES (1×P2, security)

### [P2] Redact schemeless proxy credentials before forwarding — `internal/secureenv/manager.go:348-350`
When `forward_proxy_env` is enabled and the ambient proxy uses the common **schemeless** form `user:pass@proxy.example.com:8080`, `url.Parse` treats `user` as the scheme and leaves `u.User` nil, so the redaction path returns the original value and **forwards the credentials** — defeating the redaction guarantee for users whose proxy vars omit `http://`/`https://`. Add a fallback that detects userinfo (`...@...` before the host) without a URL scheme and redacts it before appending to upstream env.

_Reviewed via `codex exec review --base origin/main` on `e72ff2f6`._

Dumbris:
Fixed in 231497a6 — thanks, this was a real leak.

`redactProxyCredentials` now handles schemeless values: when the value has no `://` but contains `@`, it re-parses with a dummy `http://` scheme so `url.Parse` recognizes the userinfo, strips it, then removes the dummy scheme. An `@` inside a path (`host/path@x`) keeps `Userinfo` nil and is left intact.

Tests added:
- `redactProxyCredentials` table: `user:pass@proxy.example.com:8080` → `proxy.example.com:8080`, `user@host:3128` → `host:3128`, schemeless-no-userinfo unchanged, at-in-path unchanged.
- End-to-end `TestBuildSecureEnvironment_ForwardsSchemelessProxyRedacted`: `HTTP_PROXY=user:pass@proxy.example.com:8080` forwarded with no `@`/`pass` in the value.

`go test ./internal/{secureenv,config,upstream}/... -race` and `golangci-lint v2.5.0` both green.

Dumbris:
## ✅ Review (Claude Code, Codex rate-limited) — ACCEPT

Rigorous substitute review while Codex is rate-limited (per maintainer direction).

**Previously-flagged P2 (schemeless credential redaction) — CONFIRMED RESOLVED.** `redactProxyCredentials` (`internal/secureenv/manager.go:340-372`) uses a two-stage approach (`url.Parse`, then a dummy-scheme re-parse for schemeless `user:pass@host:8080`). Empirically verified across 14 edge cases: IPv6 (`user:pass@[::1]:8080`), `@`-in-password, percent-encoded creds, query/fragment, empty user/pass → creds stripped, host preserved; `host/path@x`, `NO_PROXY` lists, `*`, empty → left intact.

**Opt-in correctness ✓** — `ForwardProxyEnv` defaults false; gated at `manager.go:286` + `client.go:184-188`; proxy vars explicitly excluded from the default allowlist (comment cites the #701 lesson). stdio-only. Explicit config wins (forwarding runs last, suppressed if any case-spelling already present).

**Security ✓** — no raw-value logging, no un-redacted forward path. **Tests ✓** — `proxy_forward_test.go` + `secure_env_integration_test.go` cover schemeless creds, default-off, NO_PROXY, case-alias precedence; `go test ./internal/secureenv/... ./internal/upstream/...` green (incl. `-race`); `go build ./...` exit 0.

**Non-blocking (optional hardening, not gating):** a proxy value with *surrounding whitespace* (`"  http://user:pass@host  "`) is returned unchanged because `url.Parse` fails on the leading space and the `://` guard skips the schemeless branch — a `strings.TrimSpace` at the top of `redactProxyCredentials` would close it. Extremely unlikely in practice; filed as a note, not a blocker.

**VERDICT: ACCEPT**

PR #740, opened 2026-06-20
fix(mcp): truncate direct-mode text on a UTF-8 rune boundary

electrolobzik:
## Problem

Direct-mode response truncation in `internal/server/mcp_routing.go` cuts `TextContent` at a raw byte offset — `txt = txt[:limit]` where `limit` is `ToolResponseLimit`. When the cut point lands inside a multi-byte UTF-8 sequence (emoji, accented or non-Latin text), the forwarded `TextContent` contains a partial/invalid byte sequence. Downstream JSON encoders / MCP clients then reject it or render a replacement character at the truncation boundary.

The direct-mode path intentionally uses the simple `ToolResponseLimit` byte budget rather than the caching `Truncator` (per the existing comment), so this is the right place to make the cut rune-safe.

## Fix

Add `safeTruncateBytes(s, limit)` — the largest cut `<= limit` that does not split a rune (it backs up from the byte offset to the nearest `utf8.RuneStart`) — and use it in the direct-mode forward path. ASCII behaviour is unchanged (the cut stays at `limit`).

## Test

`TestSafeTruncateBytes` covers ASCII, 2-byte and 4-byte runes at and inside the boundary, and asserts every truncated prefix is valid UTF-8 via `utf8.ValidString`.

### DIFF
--- a/internal/server/mcp_routing.go
+++ b/internal/server/mcp_routing.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/mark3labs/mcp-go/mcp"
 	mcpserver "github.com/mark3labs/mcp-go/server"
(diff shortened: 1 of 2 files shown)


Dumbris (approved):

Dumbris:
Thank you for this, @electrolob — this is a model small contribution. 🙏

A reviewer went through it carefully and the fix is correct and well-scoped:
- **Idiomatic and efficient** — using `utf8.RuneStart` for an O(1)-bounded back-scan (UTF-8 runes are ≤4 bytes) instead of converting the whole response to `[]rune` is exactly the right call for a hot truncation path; it avoids a big allocation on large response bodies.
- **Great test discipline** — the table-driven `TestSafeTruncateBytes` covers ASCII / 2-byte / 4-byte / empty / exact-boundary / zero / negative, and asserts `utf8.ValidString` on *every* result rather than hard-coded lengths — which is precisely the invariant the bug was about. The doc comment explaining *why* (downstream encoders / replacement chars) is appreciated too.

Verified locally: `go build ./...` clean, full `internal/server` package green, all 9 subtests pass, gofmt clean.

Merging now. One optional follow-up if you're ever up for it (definitely not blocking): the non-`CallToolResult` fallback path around `mcp_routing.go:276` still does a raw byte cut and has the same rune-splitting edge case — routing it through `safeTruncateBytes` too would close it fully. Thanks again!

PR #594, opened 2026-06-04
feat(registry): simplify trust/quarantine + add edit endpoint (MCP-1072)

Dumbris:
## Backend: simplify registry trust/quarantine + add edit endpoint (MCP-1072)

Foundation task that **unblocks the web (B) and macOS (C) UI tasks**. Backend-only — no `frontend/` or `native/macos/` changes (those are the B/C lanes).

### Part 1 — simplify (remove special-casing)
- **Drop quarantine-forcing for custom origins** in `add_from_registry.go`: servers added from any registry (official *or* custom) now get `Quarantined = quarantineDefault` like everything else. Provenance no longer force-quarantines.
- **Remove the config validation** that forbade `skip_quarantine=true` on custom-provenance servers (`config.go`).
- **Provenance → two plain values**: `official` / `custom` (was `official/trusted` / `custom/unverified`). Legacy persisted values are **normalized on read** (`NormalizeRegistryProvenance` applied in `initializeRegistries`) for both registry entries and per-server `source_registry_provenance`, so existing `config.db` / `mcp_config.json` keep working.
- **Fixed a latent bug**: the REST list projection hard-coded the old `"official/trusted"` literal to derive `trusted` — now normalizes + compares against `RegistryProvenanceOfficial`.
- `trusted` boolean kept as derived `official == trusted` (still consumed by UI surfaces for the third-party warning).
- Removed the "always quarantined" backend warning copy in `list_registries` (mcp.go) and the add-source help/warning text.
- Global quarantine default stays **enabled** (`IsQuarantineEnabled` nil→true, unchanged).

### Part 2 — edit endpoint + CLI
- **`PUT /api/v1/registries/{id}`** (`handleEditRegistrySource` + `EditRegistrySourceRef`): update a custom registry's name/url/servers-url. Refuses built-ins (`registry_shadows_builtin`), unknown id (`registry_not_found`), non-https url (`invalid_registry_url`), locked policy (`registries_locked`) — mirroring the add/remove-source cross-surface error pattern (#592). Changing `--url` re-derives the servers URL unless `--servers-url` is given.
- **CLI**: `mcpproxy registry edit <id> [--name … --url … --servers-url …]` (daemon-routed, like add-source) + `cliclient.EditRegistrySource`.
- Regenerated `oas/swagger.yaml` + `oas/docs.go`; updated `docs/registries.md` + `docs/api/rest-api.md`.

### Tests (TDD)
[...]

PR #745, opened 2026-06-22
feat(registries): harden registry fetches against SSRF (CWE-918, MCP-1076)

Dumbris:
## What

Hardens registry URL fetches against SSRF / CodeQL `go/request-forgery` (alerts #97/#98, MCP-1076). Registry server-list fetches use a **user-supplied** URL (`registry add-source`), so a malicious or typo'd source could point the daemon at internal/metadata endpoints (`169.254.169.254`, RFC1918 ranges).

## How

A single `isBlockedIP` predicate (new `internal/registries/ssrf.go`) is the source of truth for blocked ranges: loopback, RFC1918 private, RFC6598 CGNAT (`100.64/10`), link-local incl. the cloud-metadata endpoint, IPv6 unique-local, unspecified, multicast — fails closed on an unparseable IP. It is enforced at **three layers**:

1. **Dial time** — a `net.Dialer.Control` hook on the shared registry HTTP client checks the *actual resolved IP* before connect. This is the authoritative guard: it catches hostnames that resolve into blocked ranges and closes the DNS-rebinding TOCTOU window a parse-time check alone cannot.
2. **Fetch pre-flight** — `validateRegistryURL` rejects a literal-IP host (defense in depth + an explicit barrier at the `http.NewRequest` sink, alongside the existing scheme/host pin).
3. **add-source / edit-source** — literal-IP fail-fast (no DNS, stays pure/offline) so a bad source is refused up front with the stable `invalid_registry_url` code.

### Opt-out
New top-level `allow_private_registry_fetch` config flag (default **false** = secure). Set `true` only to run a trusted registry mirror on an internal/private address. Wired through `SetRegistriesFromConfig` so config hot-reload keeps the dial guard in sync. This is the "optional allow/deny policy" called for in the issue.

## Tests (TDD)

- `TestIsBlockedIP` — every blocked range + public allow cases incl. RFC1918/CGNAT boundaries + IPv6 + fail-closed nil.
- `TestHostLiteralBlocked`, `TestValidateRegistrySourceURL` — literal-IP rejection, hostname pass-through, opt-in bypass.
- `TestRegistryGet_BlocksLoopbackWhenGuardActive` — end-to-end: guarded fetch refused at dial time, endpoint never reached.
- `TestBuildRegistrySourceEntry_RejectsSSRFLiteralIP` — add-source fail-fast.
- The two registry-add E2E daemon tests opt in via the new config flag.

Verified locally: `go test ./internal/registries/... ./internal/config/... -race`, registry-add E2E daemon tests, add/edit-source unit tests, golangci-lint v2 (`.github/.golangci.yml`) 0 issues, gofmt, `make swagger` regenerated.

## Notes

CodeQL #97/#98 are **pre-existing** alerts on `main` (created by #572, surfaced — not introduced — by #594). This PR adds the mitigating guard; the alerts can be dismissed as mitigated once merged.

Closes MCP-1076.


Dumbris:
Changes requested — SSRF guard blocks loopback, breaking `internal/server` registry-add tests on CI (`TestCrossSurfaceConsistency_RegistryAdd`, `TestHandleUpstreamServers_AddFromRegistry_*`, `TestCrossSurfaceConsistency_OfficialClassification`). The tests use httptest servers on 127.0.0.1. Make the resolver/policy injectable (or add an opt-in allow for trusted/test hosts) so prod stays deny-by-default but tests/local pass. Details in MCP-1076.

PR #756, opened 2026-06-24
feat(index): per-profile Bleve indexes (Profiles v2 T1) — MCP-3240

Dumbris:
## Profiles v2 — T1: per-profile Bleve indexes + lifecycle

Parent: MCP-33 · Foundation that T2 (`set_profile`) wires switching to. Does **not** change URL-profile filtering yet.

### What
Introduces physically separate per-profile Bleve indexes so switching profiles is fast and a config reload that changes one profile does not re-index the others.

### `index.Manager` API
- `ForProfile(slug)` — lazily opens/creates a per-profile index, cached one-per-slug; invalid (non-filesystem-safe) slugs rejected.
- `RebuildProfileFromShared(slug, servers)` — rebuilds a profile index as a **derived view** of the shared default index, touching only that profile.
- `DropProfile(slug)` — closes + removes a profile's index directory.
- `ExistingProfileDirs` / `ClearAll` / `DeleteAll` helpers.

### On-disk layout
```
index.bleve/                 # shared default index (all servers' tools, /mcp allow-all fallback)
index.bleve/profiles/<slug>/ # one index per profile (only that profile's servers' tools)
```
**Design note (deviation from the issue's literal `index.bleve/<profile>/`):** Bleve creates its own `store/` subdirectory and metadata files directly under `index.bleve/`. A naive `index.bleve/*` enumeration for orphan cleanup mistook Bleve's `store/` dir for a profile and deleted it — caught by a real test failure during development. Namespacing under `index.bleve/profiles/` keeps per-profile indexes under `index.bleve` (the spec's spirit) while isolating them from Bleve internals so enumeration is always safe.

### Runtime wiring (`internal/runtime/profile_index.go`)
- `reconcileProfileIndexes` — after each discovery pass and config reload: build new profiles, rebuild membership-changed ones, drop removed ones (incl. orphaned dirs from a prior run); **unchanged profiles are untouched**.
- `reindexAffectedProfiles` — when a server's tools change in the shared index (discovery / disable / quarantine), refresh **only** the profiles that include that server.

The shared index stays the source of truth; per-profile indexes are derived from it (no re-discovery from upstreams).

### Tests (TDD)
- `internal/index`: `ForProfile` isolation, reload isolation (rebuild A only, B doc-count unchanged), drop removes dir, invalid-slug rejection.
- `internal/runtime`: reconcile builds + isolates, drops removed profile dir, server-scoped refresh only touches profiles with that server.

### Verification
[...]

PR #763, opened 2026-06-24
fix(index): paginate per-profile rebuild/clear beyond search-page caps (MCP-3319)

Dumbris:
## Summary

Backend robustness hardening — follow-up to the CodexReviewer `request_changes` verdict on [MCP-3307](/MCP/issues/MCP-3307) (adversarial review of per-profile Bleve indexes, PR #756 / MCP-3240). The human merged #756 pragmatically; the two correctness gaps Codex flagged are fixed here.

### Findings fixed (both shipped to `main` in `f046f905`)

1. **`GetToolsByServer` silently dropped tools beyond a 10000-hit search cap** (`internal/index/bleve.go`). Sourced by `Manager.RebuildProfileFromShared` → a server with >10k indexed tools would have its overflow omitted from the rebuilt profile index.
2. **`DeleteAll` only cleared the first 100000 docs** (`internal/index/bleve.go`). On a profile rebuild (which calls `DeleteAll` first), stale docs beyond 100k could survive, leaving the profile index inconsistent with the shared index.

### Fix

Both operations now loop over search pages until exhausted, so coverage is never bounded by a single page:

- **`GetToolsByServer`** paginates with `From`/`Size`, ending on a short final page (fewer hits than the page size).
- **`DeleteAll`** deletes page-by-page from offset `0` — each batch removes the docs just enumerated, so the next search surfaces the following page — until the index is empty (no `From` offset bookkeeping against a shrinking set).

Page size is a configurable `BleveIndex` field (default `10000`) so the regression tests exercise the pagination loop with a small, fast corpus instead of indexing >10k docs.

### Tests (TDD)

- `TestBleveIndex_GetToolsByServer_BeyondPageCap` — page size 3, 7 tools across 3 pages + an unrelated server; asserts all 7 returned and the term filter holds across pages. **Fails before the fix** (overflow dropped).
- `TestBleveIndex_DeleteAll_BeyondPageCap` — page size 3, 10 docs; asserts `GetDocumentCount()==0` after `DeleteAll`. **Fails before the fix** (docs beyond a page survive).

### Verification

- `go test ./internal/index/... -race` ✅
- `go test ./internal/runtime/ -run Profile -race` ✅ (profile rebuild/reconcile path)
- `go build ./internal/...` ✅
- `golangci-lint run --config .github/.golangci.yml ./internal/index/...` → 0 issues ✅

### Out of scope

`DeleteServerTools` carries a similar `Size = 1000` per-server cap, but it was not part of the Codex findings and isn't on the profile-rebuild path. Noting it here as a candidate follow-up; left unchanged to keep this PR scoped to the two cited findings.

### Provenance
- Codex verdict: comment `233fb835-22c7-4a44-a372-71b047eb879b` on MCP-3307
- Shipped commit: `f046f905` (PR #756)

Related #756.

### DIFF
--- a/internal/index/bleve.go
+++ b/internal/index/bleve.go
@@ -14,10 +14,19 @@ import (
 	"github.com/smart-mcp-proxy/mcpproxy-go/internal/config"
 )
 
+// defaultSearchPageSize is the page size used when paginating full-coverage
+// scans (GetToolsByServer, DeleteAll). It is a generous upper bound on the
+// number of docs in a single page; pagination loops over as many pages as
+// needed, so total coverage is never bounded by this value (MCP-3319).
+const defaultSearchPageSize = 10000
+
 // BleveIndex wraps Bleve index operations
 type BleveIndex struct {
 	index  bleve.Index
 	logger *zap.Logger
+	// searchPageSize bounds a single search page during paginated full scans.
+	// Defaults to defaultSearchPageSize; overridable in tests.
+	searchPageSize int
 }
 
 // ToolDocument represents a tool document in the index
@@ -61,8 +70,9 @@ func newBleveIndexAt(indexPath string, logger *zap.Logger) (*BleveIndex, error) 
 	}
 
 	return &BleveIndex{
-		index:  index,
-		logger: logger,
+		index:          index,
+		logger:         logger,
+		searchPageSize: defaultSearchPageSize,
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


Dumbris (commented):
ACCEPT — solid pagination fix for MCP-3319.

**What the PR does:**
- Replaces single-search-with-magic-cap in `DeleteAll` and `GetToolsByServer` with proper pagination loops.
- `DeleteAll` fetches from offset 0 each iteration after batch delete (correct — avoids stale offsets post-mutation).
- `GetToolsByServer` uses standard `from`-offset pagination for read-only iteration.
- `searchPageSize` field (default 10000) overridable in tests; enables fast regression with small corpus.

**Tests:** Both BeyondPageCap tests pass; full `internal/index` suite passes.

**No issues found.** LGTM.

PR #819, opened 2026-07-07
feat(ci): tag-blocking release QA gate — server-type matrix + invariants (Spec 081 T1)

Dumbris:
## What

Spec 081 T1 — a single tag-blocking **release qualification gate**. This PR is stage 3 (workflow + wiring); it lands on top of the stage-1/2 fixtures + driver already on this branch. It assembles the existing suites with a new server-type matrix and end-to-end invariants, produces one machine-readable verdict, and makes artifact publication depend on it: a red gate publishes nothing, mechanically.

## Architecture

Reusable workflow `.github/workflows/release-qa-gate.yml`. Every job writes a JSON **report fragment** into a shared dir; a final `verdict` job merges them against the hardcoded `internal/gatereport` manifest and exits per verdict.

| Job | Blocking manifest entries | Timeout |
|-----|---------------------------|---------|
| `build-candidate` | — (builds candidate core + fixtures + driver, one shared artifact) | 15m |
| `suite-api-e2e` | `suite/api-e2e` (`scripts/test-api-e2e.sh` unmodified) | 15m |
| `suite-race` | `suite/unit-race`, `suite/server-race` | 25m |
| `suite-scan-eval` | `suite/scan-eval` (unconditional, FR-015) | 10m |
| `matrix-invariants` | `matrix/{stdio,http,sse,docker,oauth}` + `invariant/{activity-request-id,counters,quarantine-flow,upgrade-in-place}` | 20m |
| `verdict` (`if: always()`) | merges → `gate-report.json` → `$GITHUB_STEP_SUMMARY`; exit code **is** the verdict | 10m |

`matrix` and `invariants` run in one shell step so the state-file'd core survives between them. Missing/failed blocking fragments fail the gate (fail-closed, FR-004). Every job has an explicit timeout (FR-005).

## FR coverage map

- **FR-001 / 001a** — one verdict per tag; `workflow_dispatch` dry-run against any ref (publishes nothing).
- **FR-002** — `release.yml` `release` job (and everything cascading from it) + `prerelease.yml` `release` job now `needs:` a `qa-gate` job that `uses:` the reusable workflow.
- **FR-003** — invokes `test-api-e2e.sh`, `go test -race ./internal/...`, server-edition race, and `scan-eval --gate` (not reimplementations).
- **FR-004 / 010** — fragment statuses `pass|fail|flaky|skipped|not-run|advisory-fail`; no silent skips.
- **FR-005** — explicit `timeout-minutes` per job.
- **FR-006..010** — five-type matrix (driver, stage 1/2) wired with a Docker preflight; docker cell fails-not-skips.
- **FR-011..014** — invariants wired against the live matrix instance + upgrade-in-place from the latest stable release.
- **FR-015** — scan-eval runs on every tag regardless of changed paths.
- **FR-022 / SC-004** — `cmd/release-gate/workflow_audit_test.go` parses both publishers with `yaml.v3` and asserts every artifact-publishing job's transitive `needs` includes the `qa-gate` job (statically disabled `if: false` jobs excluded). Verified it fails on an ungated `release`.

## Gating safety

Both `qa-gate` jobs share the exact trigger condition of their `release` job (release.yml: stable-only; prerelease.yml: prerelease tags only, skips the `next` branch). So a **skipped** gate skips the publish job too — it can never silently un-gate. `workflow_dispatch` of `release.yml` is unchanged (all jobs still guard on the tag ref). RC tags don't double-run the gate in `release.yml` (prerelease.yml owns the RC gate).

## Notable decisions

- The bare FR-003 `scan-eval` command omits the **required** `--corpus` flag; I use the full `eval.yml` invocation (`--corpus …/detect_corpus_v1.json --gate --min-recall 0.90 --max-fp 0.05`) so the gate actually runs.
- **FR-003 `-short` fallback**: `suite-race` runs the full non-short suites; if that empirically blows the 25m budget, fall back to `-short` in that job only with a loud comment — the heavy property/timing tests still run unguarded in `e2e-tests.yml`'s `stress-tests` job.
- **FR-011 correlation finding**: the activity-log invariant found that a caller-supplied `X-Request-Id` is **not** persisted on `tool_call` activity records today (`internal/server/mcp.go` synthesizes its own per-call ids). The check falls back to locating each call by a unique argument **nonce** and proving the core-recorded request id round-trips via `?request_id=`, and records the limitation in the fragment rather than hiding it. Persisting the caller's header end-to-end is a middleware change left out of this stage.

## T2–T4 extension points

Manifest reserves three non-blocking slots (recorded `not-run`/`not-implemented-yet`): `reserved/web-ui-sweep` (T2 Playwright sweep vs embedded frontend), `reserved/macos-smoke` (T3 advisory tray smoke; promotion criterion — 3 consecutive clean tags — documented), `reserved/surface-consistency` (T4 REST/CLI/Web-UI/tray agreement). See `docs/development/release-gate.md`.

## Dry-run instructions

`gh workflow run release-qa-gate.yml -f ref=<branch-or-sha>` (or Actions UI → Release QA Gate → Run workflow). A dry run publishes nothing and never qualifies a later tag.

> Note: `release-qa-gate.yml` carries a **TEMP** `push:` trigger on `081-release-qa-gate-t1` (commented `# TEMP: remove before merge`) so the branch gets real CI; remove before merge.

## Testing

- `go build ./...`, `go vet`, gofmt clean.
- `golangci-lint run --config .github/.golangci.yml ./cmd/release-gate/...` → 0 issues.
- `go test -race ./cmd/release-gate/` passes (incl. the new audit test).
- `actionlint` clean on the new workflow; no new findings on the edited publishers (pre-existing shellcheck infos untouched).
- Python `yaml.safe_load` parses all three workflows.

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -757,8 +757,19 @@ jobs:
           name: installers-${{ matrix.goos }}-${{ matrix.goarch }}
           path: installers-artifact/*
 
+  # Spec 081: release qualification gate (same reusable workflow as release.yml).
+  # Guarded to prerelease TAG refs only — skipped on `next` branch pushes,
+  # consistent with the `release` job's own tag condition, so branch CI is not
+  # burdened with the full gate. A red gate blocks the prerelease publish
+  # (FR-002); a skipped gate (branch push) skips `release` too (never un-gated).
+  qa-gate:
+    uses: ./.github/workflows/release-qa-gate.yml
+    permissions:
+      contents: read
+    if: startsWith(github.ref, 'refs/tags/v') && (contains(github.ref, '-rc.') || contains(github.ref, '-next.'))
+
   release:
-    needs: build
+    needs: [build, qa-gate]
     runs-on: ubuntu-latest
     environment: staging
     # Only create releases for tag pushes, not branch pushes
--- a/.github/workflows/release-qa-gate.yml
+++ b/.github/workflows/release-qa-gate.yml
@@ -1,0 +1,419 @@
+name: Release QA Gate
+
+# Spec 081 T1 — Release Qualification Gate.
+#
+# One reusable workflow that assembles the existing suites (test-api-e2e.sh,
+# go race tests, scan-eval --gate) with the new server-type matrix
+# (stdio/http/sse/docker/oauth: connect, list, call, reconnect) and the US2
+# invariant checks (activity-log request-id correlation, counter deltas,
+# quarantine end-to-end, upgrade-in-place). It produces ONE machine-readable
+# verdict (gate-report.json) via internal/gatereport; the `verdict` job exits
+# per that verdict so a caller (`release.yml` / `prerelease.yml`) that lists
+# this workflow in a publish job's `needs:` publishes nothing on a red gate
+# (FR-001, FR-002, FR-004).
+#
+# Triggered as a reusable workflow (workflow_call) by the publishers, or
+# manually against any ref (workflow_dispatch — a dry run that publishes
+# nothing and never counts as qualification for a later tag, FR-001a).
+#
+# T2/T3/T4 extension slots (reserved manifest entries, recorded as
+# not-run/"not-implemented-yet" until their stage lands):
+#   reserved/web-ui-sweep        — T2 Playwright sweep against the candidate
+#   reserved/macos-smoke         — T3 macOS tray smoke (advisory, FR-019/020)
+#   reserved/surface-consistency — T4 REST/CLI/Web-UI/tray state agreement
+# See docs/development/release-gate.md.
+
+on:
+  workflow_call:
+    inputs:
+      ref:
+        description: "Ref to qualify (defaults to the caller's triggering commit)"
+        required: false
+        type: string
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Ref to qualify (dry run — publishes nothing, FR-001a)"
+        required: false
+        type: string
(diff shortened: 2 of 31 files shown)


Dumbris:
**ReleaseEngineer review (MCP-4092 fallback — GeminiCritic adapter failed, reviewed directly)**

**VERDICT: ACCEPT**

### Verification

- `go build ./cmd/release-gate/ ./cmd/mcpfixture/` ✓
- `go test ./cmd/release-gate/ ./internal/gatereport/` ✓ (both pass)
- `go vet ./cmd/release-gate/ ./internal/gatereport/` ✓
- `actionlint .github/workflows/release-qa-gate.yml` ✓ (zero findings)
- `TestPublishJobsGatedOnQAGate` passes — FR-022/SC-004 audit confirmed: every publishing job in both `release.yml` and `prerelease.yml` transitively depends on `qa-gate`

### Architecture check

Fail-closed wiring is correct:
- `verdict` job runs `if: always()`; its exit code is the gate verdict the publishers depend on
- Missing blocking fragments fail the merger explicitly (FR-004, no silent skips)
- Gate condition in both publishers exactly matches their `release` job guard — skipped gate = skipped publish, never un-gated
- `continue-on-error: true` on fragment downloads in `verdict` is correct (no-fragments → merger hard-fails)

### One pre-merge action

**Remove the TEMP `push:` trigger** (lines 39–43 of `release-qa-gate.yml`) before merging. It is already marked `# TEMP: remove before merge`. No other changes needed.

PR #822, opened 2026-07-09
fix(packaging): allow web-UI config saves on linux by adding /etc/mcpproxy to ReadWritePaths

Eun:
# Pull Request

## Description
When saving the configuration, the backend creates a new file at `/etc/mcpproxy/mcp_config.json.tmp.<suffix>`. However, when systemd starts the process, it restricts write access to the `/etc/mcpproxy` directory. This pull request modifies the service file to enable write permissions for the `/etc/mcpproxy` directory.

## Testing
- [ ] I have added/updated tests that prove my fix is effective or my feature works
- [ ] All existing tests pass

### DIFF
--- a/packaging/linux/mcpproxy.service
+++ b/packaging/linux/mcpproxy.service
@@ -23,7 +23,7 @@ ProtectKernelModules=true
 ProtectControlGroups=true
 RestrictSUIDSGID=true
 LockPersonality=true
-ReadWritePaths=/var/lib/mcpproxy
+ReadWritePaths=/var/lib/mcpproxy /etc/mcpproxy
 
 # Allow binding to privileged ports if user reconfigures listen address
 AmbientCapabilities=CAP_NET_BIND_SERVICE


Dumbris:
Thank you for tracking this down, @Eun — the diagnosis is exactly right. `ProtectSystem=strict` mounts `/etc` read-only, and `SaveConfig` writes `mcp_config.json.tmp.<hex>` next to the config before renaming it, so the save fails inside the sandbox. Adding `/etc/mcpproxy` to `ReadWritePaths=` is the correct minimal fix. 🙏

One thing your patch can't cover from the unit file alone: `ReadWritePaths=` lifts systemd's namespace restriction, but POSIX permissions still apply. `/etc/mcpproxy` is an implicit parent directory in `nfpm.yaml`, so a fresh `.deb`/`.rpm` creates it `root:root 0755` — and the service runs as `User=mcpproxy`, which then can't create the temp file in it. On your host it presumably already had workable ownership, which is why your local test passed.

So I'm merging this as-is and landing a companion commit that makes `postinstall.sh` do `install -d -m 2770 -o root -g mcpproxy /etc/mcpproxy` (the same idiom already used for `/var/lib/mcpproxy`), which also repairs existing installs on upgrade. Both will ship in v0.48.1.

Note for maintainers: `auto-pass-qa-gate` can't post a commit status on fork PRs (`403 Resource not accessible by integration`), so the required `qa-gate` context was attested manually on `b5007909`.

Dumbris (approved):
Correct diagnosis and the right minimal fix for the systemd sandbox. Merging with a companion commit for the /etc/mcpproxy directory permissions.

PR #857, opened 2026-07-15
fix(runtime): add config file watcher so external edits hot-reload (HOTRELOAD)

Dumbris:
## Problem

Editing `mcp_config.json` externally (editor save, `jq > tmp && mv`, CLI commands against a running core) never took effect until restart, even though CLAUDE.md and the swagger docs both claim "the file watcher hot-reloads". Only REST `PATCH /api/v1/config` worked, because it goes through a different path (`ApplyConfig`). Found by v0.51.0-rc.1 QA (2026-07-15).

## Root cause

The watcher never existed. `fsnotify` was only an *indirect* dependency (`go.mod:89`) with zero imports; the code itself admitted it (`internal/runtime/lifecycle.go:1117`, `:1426`, `cmd/mcpproxy/telemetry_cmd.go:294` — "there is no fsnotify watcher"). The complete disk-reload path `Runtime.ReloadConfiguration()` (`internal/runtime/lifecycle.go:1081`) existed with no automatic caller, and `configsvc` was built watcher-ready (`Update.Source` example "file_watcher", `configsvc/service.go:29`) but nothing ever fed it.

## Fix

New `internal/runtime/config_watcher.go`, started from `StartBackgroundInitialization` and tied to the runtime app context:

- Watches the config file's **parent directory** (never the file), so rename/recreate writes — `mv tmp file`, atomic-write editors, mcpproxy's own `atomicWriteFile` — can never orphan the watch. Covers in-place truncate writes too.
- Debounces events 500 ms and funnels into the canonical disk-reload path `Runtime.ReloadConfiguration()` (same events / telemetry `NotifyConfigChanged` / update-check re-gating / ConnectAll + reindex as manual reload).
- **Self-write suppression**: disk bytes are compared against the marshaled in-memory snapshot, so mcpproxy's own `ApplyConfig`/`SaveConfiguration` saves never echo back into a redundant reload. No loop risk: `ReloadConfiguration` never writes the file.
- Invalid JSON keeps the previous config (`ReloadFromFile` fails before mutating state); the watcher survives and picks up the next valid write.
- Watcher startup failure degrades gracefully (warning; server keeps running without hot-reload).

Also: promoted `fsnotify v1.9.0` indirect → direct (already in go.sum, no new third-party code), updated the three stale "there is no fsnotify watcher" comments, and documented debounce / invalid-JSON / restart-required semantics in `docs/configuration.md`.

Known limitations (accepted): a config-path change after startup is not re-watched; restart-required fields (`listen`, `data_dir`) edited externally reload into memory but still need a restart — same semantics as the manual reload path.

## Tests

New `internal/runtime/config_watcher_test.go` (written first, TDD — failed with `startConfigFileWatcher undefined` pre-fix):

- `TestConfigWatcher_RenameWriteTriggersReload` — atomic tmp+rename edit hot-reloads
- `TestConfigWatcher_TruncateWriteTriggersReload` — in-place `>` edit hot-reloads
- `TestConfigWatcher_InvalidJSONKeepsOldConfigThenRecovers` — bad JSON keeps old config + version; watcher recovers
- `TestConfigWatcher_SelfWriteSuppressed` — `ApplyConfig`'s own save produces no `file_reload` echo
- `TestConfigWatcher_DebounceCoalesces` — 10 rapid writes collapse to ≤3 reloads, final value lands
- `TestConfigWatcher_MissingDirGracefulDegradation` — missing parent dir errors without panic

Commands run:
- `go test -race ./internal/runtime/...` — all pass
- `go test -race ./cmd/mcpproxy/...` — pass
- `go test -tags server ./internal/serveredition/... -race` — pass
- `go build ./cmd/mcpproxy` and `go build -tags server ./cmd/mcpproxy` — both compile
- `golangci-lint run --config .github/.golangci.yml ./internal/runtime/... ./cmd/mcpproxy/...` — 0 issues

### DIFF
--- a/cmd/mcpproxy/telemetry_cmd.go
+++ b/cmd/mcpproxy/telemetry_cmd.go
@@ -290,9 +290,11 @@ func runTelemetryDisable(cmd *cobra.Command, _ []string) error {
 	// command never appears to hang on the (best-effort) beacon below.
 	fmt.Println("Telemetry disabled.")
 
-	// One-time opt-out beacon. When a daemon is running it does NOT auto-reload
-	// this file (there is no fsnotify watcher), so the CLI is responsible for the
-	// beacon in the CLI-driven path. Route it through the SAME guarded server-side
+	// One-time opt-out beacon. A running daemon now hot-reloads this file via
+	// the fsnotify config watcher (config_watcher.go) and fires its own beacon
+	// through NotifyConfigChanged — but no daemon may be running, so the CLI
+	// stays responsible for the beacon in the CLI-driven path. Route it
+	// through the SAME guarded server-side
 	// entry point (EmitOptOutBeacon applies the dev-build/semver, env, and
 	// anon-id guards and owns the single send) rather than duplicating the send
 	// or bypassing a guard. A short timeout keeps this from blocking on a slow
--- a/cmd/mcpproxy/telemetry_cmd.go
+++ b/cmd/mcpproxy/telemetry_cmd.go
@@ -290,35 +290,27 @@ func runTelemetryDisable(cmd *cobra.Command, _ []string) error {
 	// command never appears to hang on the (best-effort) beacon below.
 	fmt.Println("Telemetry disabled.")
 
-	// One-time opt-out beacon. A running daemon hot-reloads this file via the
-	// fsnotify config watcher (config_watcher.go) and fires its own beacon
-	// through NotifyConfigChanged — its once-latch is process-local, so if the
-	// CLI also sent one the beacon would go out TWICE. When a daemon is
-	// detected (same socket check as the other daemon-aware commands), the CLI
-	// therefore delegates the send to it. With no daemon running the CLI stays
-	// responsible, routing through the SAME guarded server-side entry point
+	// One-time opt-out beacon. The CLI ALWAYS sends it after persisting the
+	// disable, routing through the SAME guarded server-side entry point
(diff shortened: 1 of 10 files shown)


Dumbris:
Review provenance for the merger:

- 7 Codex cross-model review rounds were run on this PR (rounds 6 and 7 beyond the standard 5-round cap were explicitly human-authorized).
- 9 findings were verified and fixed across rounds 1-7, most recently: stale self-write marker on failed save, telemetry opt-out beacon double-send (resolved by reverting delegation and documenting the accepted double-send trade-off), disk-reload parity for live component side effects (logging, truncator, observability cadence), and the single-slot self-write marker race.
- Round 7 left two findings, resolved as follows:
  - Finding 1 (single-slot `lastSelfWrite` marker race): fixed in c9a369ab — the marker is now a bounded recent-self-writes set (last 4 payloads, 10s TTL) with all prior semantics preserved; regression test `TestConfigWatcher_BackToBackSelfWritesBothSuppressed` reproduces the race.
  - Finding 2 (truncator swap not reaching the already-constructed MCP handler): pre-existing on the ApplyConfig/API path, not introduced by this PR; tracked in #861 together with the broader config dual-path parity work.
- Verification at c9a369ab: both-edition builds, `go test -race` on `internal/runtime` + `cmd/mcpproxy`, golangci-lint v2 strict config (0 issues), API E2E suite 65/65 green; qa-gate posted success.

Issue #306, opened 2026-02-16
retrieve_tools: annotations/call_with always defaults to call_tool_read due to tool name parsing bug


Mikhail164th:
## Bug

`handleRetrieveTools` in `internal/server/mcp.go:851` tries to extract server name from tool name by splitting on `:`:

```go
parts := strings.SplitN(result.Tool.Name, ":", 2)
if len(parts) == 2 {
    annotations := p.lookupToolAnnotations(parts[0], parts[1])
```

But `result.Tool.Name` is just the tool name (e.g. `"browser_close_tab"`), not `"server:tool"`. The server name is stored separately in `result.Tool.ServerName`.

Since tool names don't contain `:`, `len(parts)` is always 1, so the code falls through to the else branch:

```go
} else {
    mcpTool["call_with"] = contracts.ToolVariantRead // Default to read
}
```

This means `call_with` is **always** `"call_tool_read"` for every tool, regardless of annotations.

## Verified

- mcpproxy v0.17.5
- Trace log confirms annotations ARE parsed from upstream (`"Tool with annotations from server"` for 5 tools with `destructiveHint`/`readOnlyHint`)
- `lookupToolAnnotations(serverName, toolName)` works correctly when called with proper args
- `DeriveCallWith()` works correctly
- The only break is the `SplitN(result.Tool.Name, ":", 2)` on line 851

## Fix

Replace lines 851-861:

```go
// Before (broken):
parts := strings.SplitN(result.Tool.Name, ":", 2)
if len(parts) == 2 {
    annotations := p.lookupToolAnnotations(parts[0], parts[1])
    ...
} else {
    mcpTool["call_with"] = contracts.ToolVariantRead
}

// After (fixed):
annotations := p.lookupToolAnnotations(result.Tool.ServerName, result.Tool.Name)
if annotations != nil {
    mcpTool["annotations"] = annotations
}
mcpTool["call_with"] = contracts.DeriveCallWith(annotations)
```

## Impact

All tools from all servers get `call_with: "call_tool_read"` regardless of their annotations. Destructive tools (close, delete, etc.) are not flagged as destructive in retrieve_tools response.

PR #874, opened 2026-07-21
fix(server): canonicalize retrieve_tools identity to server:tool

### LINKED ISSUE #871: retrieve_tools can return bare tool ids that describe_tool/call_tool_* cannot use
### Summary

`retrieve_tools` can return a bare tool name, while `describe_tool` and `call_tool_read/write/destructive` require the `<server>:<tool>` format. That makes the documented progressive-disclosure flow break for index entries that have `ToolMetadata.Name` as the short tool name and `ToolMetadata.ServerName` set separately.

### Why this matters

The tool descriptions for `call_tool_*` say to use the exact name returned by `retrieve_tools`. `describe_tool` also says its `tool_ids` must come from `retrieve_tools` in `<server>:<tool>` form.

When `retrieve_tools` returns `create_issue` instead of `github:create_issue`, the agent follows the instructions but the next step fails with an invalid/not-found tool id.

### Related existing work

I found related but distinct prior work:

- #306 / #307 fixed annotation/call_with parsing when names may or may not include a `server:` prefix.
- #853 added compact router / `describe_tool`, which makes this contract more visible.

I did not find an existing issue specifically covering `retrieve_tools` returning bare ids that are not directly usable by `describe_tool` or `call_tool_*`.

### Expected behavior

When `ToolMetadata.ServerName` is known, `retrieve_tools` should return the canonical `<server>:<tool>` identity in:

- compact mode `id`
- full mode `name`
- disabled/locked entries where applicable

### Notes

I have a small patch with tests that canonicalizes the response identity without changing the index storage format.

Dumbris:
Fixes #871

`retrieve_tools` returned the bare indexed tool name (`echo`) while `describe_tool`/`call_tool_*` require `server:tool` — breaking the spec-085 progressive-disclosure round-trip for every live-indexed tool (compact mode was unrecoverable: no server field).

## Change
- Canonicalize identity at the index **read** seams (`SearchTools`, `GetToolsByServer` in `internal/index/bleve.go`) via `canonicalToolName` with a double-prefix guard. Stored docs unchanged — works on existing indexes.
- Companion fix: `lookupToolAnnotations` now normalizes its input (central fix in `mcp_visibility.go`), so annotations/`call_with` derivation keeps working with canonical names — without this, canonicalization would silently regress #306 (annotations dropped, `call_with` degrading to `call_tool_read`).
- Side effects now correct: `include_stats` usage lookup keys match how usage is recorded; disabled/locked entry names consistent with the quarantine pass.

## Verification
- Live app (rebuilt binary, `server-everything` upstream): compact id and full name now `everything:echo`; `describe_tool`/`call_tool_read` accept the exact returned id; annotations/`call_with` verified intact across 6 tools; `include_stats` populated.
- New tests: bleve canonicalization + double-prefix guard, retrieve→describe round-trip, annotation live-path regression test. `go test -race` green, 65/65 API E2E green, golangci-lint v2 clean.

Note: while updating `TestE2E_QuarantineConfigApply` we noticed a pre-existing gap (not introduced here): `/api/v1/index/search` applies no quarantine filtering. Will file separately.

### DIFF
--- a/bench/armindex.go
+++ b/bench/armindex.go
@@ -69,7 +69,7 @@ func (ai *ArmIndex) SearchFunc() SearchFunc {
 		}
 		ranked := make([]string, 0, len(results))
 		for _, r := range results {
-			ranked = append(ranked, r.Tool.ServerName+":"+r.Tool.Name)
+			ranked = append(ranked, index.CanonicalToolName(r.Tool.ServerName, r.Tool.Name))
 		}
 		return ranked, nil
 	}
--- a/bench/live.go
+++ b/bench/live.go
@@ -10,6 +10,8 @@ import (
 	"os"
 	"strconv"
 	"time"
+
+	"github.com/smart-mcp-proxy/mcpproxy-go/internal/index"
 )
 
 // LiveClient talks to a running mcpproxy instance (e.g. the bench
@@ -159,7 +161,10 @@ func (c *LiveClient) Search(ctx context.Context, query string, limit int) (ranke
 	}
 	ranked = make([]string, 0, len(resp.Results))
 	for _, r := range resp.Results {
-		ranked = append(ranked, r.Tool.ServerName+":"+r.Tool.Name)
+		// The search endpoint returns canonical "server:tool" names since #871;
+		// CanonicalToolName keeps compatibility with older proxies that return
+		// the bare tool name.
+		ranked = append(ranked, index.CanonicalToolName(r.Tool.ServerName, r.Tool.Name))
 	}
 	return ranked, latency, nil
 }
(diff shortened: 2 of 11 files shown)


Dumbris:
Cross-model review (Codex 0.144.6 / gpt-5.6-sol): **ACCEPT** in round 1, no findings. CI: all checks green except `Analyze (go)` — the CodeQL 4.37.0/4.37.1 infra mismatch that's failing on main itself (not a required check). Extra hardening while getting CI green: the REST `/api/v1/index/search` response keeps its bare-name contract (`name` + `server_name`) since the external mcp-eval D1 gate assembles ids itself — canonical `server:tool` ids are an MCP-surface contract only. D1 retrieval gate back to baseline recall.

PR #875, opened 2026-07-21
fix(security): redact env secrets and URL query params in server listings

### LINKED ISSUE #872: upstream_servers list should redact env secrets and sensitive URL query params
### Summary

`upstream_servers` with `operation=list` already redacts sensitive header values, but it still returns:

- `url` with sensitive query params in clear text
- `env` values in clear text

This can expose API keys, bearer tokens, OAuth callback tokens, or other credentials to an MCP caller that only needs operational server metadata.

### Related existing work

Header redaction was already fixed in #425, so this is not a duplicate of that PR. This issue is limited to the remaining secret-bearing fields in the MCP `upstream_servers list` response.

### Expected behavior

By default:

- Sensitive URL query params such as `token`, `access_token`, `client_secret`, etc. should be redacted.
- Secret-looking env keys such as `API_KEY`, `TOKEN`, `CLIENT_SECRET`, `PASSWORD`, etc. should be masked.
- Config references such as `${env:NAME}` and `${keyring:NAME}` should remain visible as references, not secrets.
- Non-sensitive env values should remain readable.

### Notes

I have a small patch that reuses the existing OAuth redaction helpers and mask format from #425.

Dumbris:
Fixes #872

Server listings masked header values (#425) but returned env values and URL query secrets (`?access_token=…&client_secret=…`) in clear text on all three parity seams: MCP `upstream_servers list`, REST `GET /api/v1/servers`, and the SSE `servers.changed` payload. Repro also showed the raw URL echoing into `connection_status.last_error`/`health.detail` on connect failures.

## Change
- New helpers next to the #425 code in `internal/oauth/logging.go`:
  - `RedactEnvValues` — masks values of secret-looking env keys (`TOKEN`/`SECRET`/`KEY`/`PASSWORD`/…) as `••••<last2> (<N> chars)`; `${env:…}`/`${keyring:…}` references stay visible; non-sensitive keys (e.g. `LOG_LEVEL`) stay readable; nil-in/nil-out.
  - `RedactURLQueryParams` — `net/url`-based (not the regex `RedactURL`, which would mask reference values); masks sensitive query param values, keeps references visible, preserves URL validity.
- Applied at all three seams, gated by the existing `reveal_secret_headers` flag; also scrubs URL secrets from `last_error`/`health.detail`/`diagnostic.cause`.
- Follow-up hardening from review: per-server diagnostics endpoint and `/api/v1/diagnostics` + doctor no longer leak the raw connect error; **macOS tray `saveEdits` no longer sends the full (masked) env map back** — it now diffs like headers, which also fixes a latent bug where an unrelated tray edit could persist masked values over real secrets.
- Write paths untouched: PATCH deep-merge means diff-sending clients never round-trip masked values (verified live).

## Verification
- Live app: headers masked as before; URL query values masked with valid URL structure; `API_KEY`/`MY_TOKEN` masked, `LOG_LEVEL` readable, `${env:HOME}` shown as reference; zero plaintext-secret hits grepping full MCP+REST responses; PATCH round-trip left the on-disk secret intact.
- `go test -race` green across oauth/httpapi/runtime/server (three stale plaintext-env assertions updated to the masked contract), 65/65 API E2E green, lint clean.


Dumbris:
Cross-model review (Codex): clean after 4 rounds. R1: masked-URL round-trip on the tray write path + heuristic gaps (DSN-style env vars, param-name variants, unanchored ${env:} refs). R2 caught a high in the new unmask helper — restoring a stored secret onto a changed host — now authority-safe and component-wise. R3: malformed-URL mask persistence edge case. R4: **ACCEPT**. CI green except the main-broken CodeQL canary (`Analyze (go)`, not required).

PR #876, opened 2026-07-21
fix(runtime): index approved tools immediately after approval + upstream refresh op

### LINKED ISSUE #873: Approved tools should become discoverable without another discovery pass
### Summary

After approving pending tools with `approve_tool` or `approve_all_tools`, the approval records are updated, but the approved tool snapshots are not immediately added back to the search index. In practice, an operator may need another discovery/reindex cycle before `retrieve_tools` and `describe_tool` can see the approved tools.

### Why this matters

The approval action is the moment an operator says the current tool description/schema is trusted. After that succeeds, the current approved snapshot should be discoverable without requiring a separate upstream refresh.

### Related existing work

I found related but distinct prior work:

- #209 / #212 implemented automatic reindexing from `notifications/tools/list_changed`.
- #608 / #609 added configurable discovery/health-check timing.

Those cover upstream-driven or scheduled discovery. This issue is specifically about the local approval workflow and an explicit operator recovery path.

### Expected behavior

When `approve_tool` or `approve_all_tools` approves a current snapshot:

- The approved snapshot should be indexed immediately when the server is enabled and not server-quarantined.
- Disabled tools and tools denied by static config should remain hidden.
- Profile indexes and compact signature cache should stay consistent with the shared index.
- Operators should also have an explicit per-upstream `refresh`/`reindex` operation for recovery when the discovery state is stale.

### Notes

I have a patch with tests that:

- indexes approved snapshots after `ApproveTools`
- covers `ApproveAllTools`
- keeps server-level quarantined tools withheld
- adds `upstream_servers operation=refresh|reindex` for one upstream

Dumbris:
Fixes #873

`ApproveTools`/`ApproveAllTools` only wrote approval records — indexing happens exclusively in `applyDifferentialToolUpdate`, reached only from discovery. So a just-approved tool was immediately *callable* but invisible to `retrieve_tools`/`describe_tool` until the next sweep (default 5 min; indefinitely with sweeps disabled). Reproduced live with a rug-pull test server.

## Change
- `reindexServerToolsAfterApprovalChange` (async, `AppContext`-scoped) runs after `ApproveTools`/`BlockTools`/`SetToolEnabled` changes: takes the last-good tool snapshot (now also populated by the single-server discovery path) and re-runs `applyDifferentialToolUpdate` — so profile indexes and the spec-085 signature cache stay consistent for free, and block/disable now *removes* tools promptly too.
- **Security guard** (mandatory): the reindex bails unless the server is enabled and not quarantined — `applyDifferentialToolUpdate` itself never checks server quarantine, so an unguarded call would index a quarantined server's withheld tool descriptions. `DiscoverAndIndexToolsForServer` got the same guard (review found the new `refresh` op could otherwise index a quarantined server, and was reachable by non-admin agent tokens — both closed).
- Operator recovery (issue's ask): new `upstream_servers` operation `refresh` (pure rediscover+reindex, no security-state change) + REST `POST /api/v1/servers/{id}/refresh`; the operation enum also now advertises the previously-undocumented `enable`/`disable`/`restart`.
- Spec-085 golden snapshot updated surgically for the enum change (delta assertions preserved).

## Verification
- Live app (rug-pull server, `tool_discovery_interval=30m` so no sweep could mask the result): pending tool invisible → `approve_all_tools` → visible to `retrieve_tools`/`describe_tool` within seconds; `refresh` op works via MCP and REST; quarantined-server probe confirms tools stay withheld.
- New tests: approve→indexed, block→removed, disable→removed, quarantined-server-not-indexed (security), full quarantine-flow reindex. `go test -race` green, API E2E green, lint clean.


Dumbris:
Cross-model review (Codex): clean after 3 rounds. R1: REST `/refresh` agent-token gate, quarantine TOCTOU, zero-tool refresh semantics. R2: eligibility guard on both full-sweep index writes (closing the quarantine→sweep-restores-tools path) and the same admin gate on `/discover-tools`. R3: **ACCEPT**. CI green except the main-broken CodeQL canary (`Analyze (go)`, not required). Note: other pre-existing REST mutating endpoints (enable/disable/restart/quarantine) also lack agent-token gating — deserves a separate consistency pass.

PR #899, opened 2026-07-23
fix(server): trusted_hosts allowlist for reverse-proxy Host validation

### LINKED ISSUE #898: [Bug]: Host header validation breaks reverse proxy setups without explicit override (403 Forbidden: invalid Host header)
## Description

When running mcpproxy behind an nginx reverse proxy (via CloudPanel), requests are rejected with:Forbidden: invalid Host header

This happens even though the proxy correctly forwards the request to mcpproxy's
listen address. It only works if I explicitly override the `Host` header in nginx
to match mcpproxy's bind address (e.g. `127.0.0.1:8004`) instead of passing through
the original public domain.

## Environment

- mcpproxy version: v0.51.0 (also reproduced on v0.52.0-rc.2)
- Listen address: `127.0.0.1:8004`
- Reverse proxy: nginx (via CloudPanel)
- A separate instance running **v0.49.0** on the same host/port, with a plain
  nginx config (no Host override, no buffering changes), works fine — same
  auth method (`?api_key=` query param).

## Workaround

Forcing the Host header to match the local bind address resolves it:

```nginx
proxy_set_header Host 127.0.0.1:8004;
proxy_buffering off;
proxy_request_buffering off;
```

## Questions

1. Is this an intentional DNS-rebinding protection added between v0.49.0 and
   v0.51.0? I couldn't find it documented in the release notes or CHANGELOG.md
   for either version.
2. If intentional, is there a config option (e.g. `trusted_hosts` /
   `allowed_hosts`) to allow specific public domains without having to spoof
   the Host header at the proxy layer? Overriding Host to the loopback address
   feels like a workaround rather than the intended setup for reverse-proxied
   deployments.
3. If this is documented somewhere, a pointer to the docs page for reverse
   proxy / Host validation config would be appreciated — I couldn't find one
   at docs.mcpproxy.app.

Happy to provide full nginx configs (before/after) and logs if useful.

## Expected Behavior

Requests forwarded through a reverse proxy with the original public domain in
the `Host` header (e.g. `Host: mcp.example.com`) should be accepted by mcpproxy,
either:

- by default (if Host validation isn't intended to block legitimate
  reverse-proxied setups), or
- after configuring the public domain as a trusted/allowed host via a config
  option (e.g. `trusted_hosts` in `mcp_config.json` or an environment
  variable), without needing to override the `Host` header to the internal
  loopback address (`127.0.0.1:8004`) at the nginx layer.

Currently, the only way to avoid the `403 Forbidden: invalid Host header`
error is to spoof the Host header to match mcpproxy's bind address, which
works but isn't documented anywhere and feels like an unintended workaround
rather than the supported configuration path for reverse-proxied deployments.

## Steps to Reproduce

1. Run mcpproxy with `--listen 127.0.0.1:8004`
2. Reverse proxy `mcp.example.com` → `127.0.0.1:8004` via nginx with standard
   `proxy_set_header Host $host;`
3. Send a request to `https://mcp.example.com/mcp?api_key=...`
4. Response: `403 Forbidden: invalid Host header "mcp.example.com"`

### MCPProxy version

 v0.51.0 

### Operating system

Linux

### Relevant logs

```shell

```

### Pre-submission checklist

- [x] I searched existing issues and did not find a duplicate
- [x] I ran `mcpproxy doctor` and reviewed the output

Dumbris:
Fixes #898

## Problem

Since v0.50 (mcp-go v0.55.1 → v0.56.0 bump in #828), running mcpproxy behind a reverse proxy fails with `403 Forbidden: invalid Host header`. mcp-go v0.56.0 added DNS-rebinding protection — requests on a loopback listener must carry a loopback `Host` header — and mcpproxy exposed no escape hatch. The only workaround was spoofing `Host` to the bind address at the proxy layer.

## Fix

- **`trusted_hosts` config key** (env `MCPPROXY_TRUSTED_HOSTS`, hot-reloadable): non-loopback `Host` values accepted on loopback listeners. Case-insensitive; entry without port matches any port, with port exact; leading-dot entry (`.example.com`) is a subdomain wildcard (Django/Vite convention); `"*"` disables validation (documented as a DNS-rebinding footgun). Default empty = exact pre-existing protection, no security regression.
- mcp-go's built-in check disabled on all five `StreamableHTTPServer` mounts, replaced by `hostValidationMiddleware` with identical default semantics (non-loopback listeners and unix-socket/tray connections never rejected).
- **Origin validation** (MCP spec MUST): a present `Origin` on a loopback listener must be a well-formed http(s)/ws(s) origin with a loopback or trusted host, else 403. Absent `Origin` (all non-browser clients) always passes — cannot re-break proxied deployments.
- Trailing-dot FQDNs (`example.com.`) match their undotted entries.
- **Docs**: new **Operations → Reverse Proxy Deployment** guide (nginx + Caddy, auth guidance) wired into the docs.mcpproxy.app sidebar — the published site previously had no reachable page on this (the reporter's exact complaint) — plus config-file/env-var reference updates and a disambiguation note on the 403 error page.

## Design validation

A 105-agent deep-research pass over the ecosystem confirmed the allowlist is the idiomatic solution (MCP Python SDK `allowed_hosts`, TS SDK `allowedHosts`, Blockscout, Vite/webpack/Django) and that default-empty matches the secure-by-default norm (CVE-2025-66414/66416, CVE-2026-34742). Both Go SDKs offer only a binary disable — this is strictly more granular.

## Verification

- Unit: 40+ case matrix (Host/Origin, IPv6, ports, case, wildcards, trailing dots, unix-socket path) + config load/env override + hot-reload change detection.
- Local E2E: exact issue scenario 403→200 with `trusted_hosts`, untrusted hosts/origins still 403, hot-reload live.
- `./scripts/test-api-e2e.sh` green; `go test -race` (server/config/runtime) green; golangci-lint v2 clean; swagger regenerated; Docusaurus build clean with broken-link check.
- **Codex cross-model review: 5 rounds, final verdict LGTM** (fixed along the way: hot-reload change reporting, origin shape strictness, trailing-dot FQDNs, port validation, docs precision).


Dumbris:
CI note: the two red `Analyze (go/javascript-typescript)` jobs fail with a CodeQL infra error — `Loaded a configuration file for version '4.37.3', but running version '4.37.0/1'` (bundle/tool-cache version mismatch, fails in ~20s before any analysis). The same jobs are failing on the last 3 pushes to `main` (e.g. run 29892885137), so this is pre-existing and unrelated to this change. All other checks are green, including E2E, race tests, and swagger-verify.

PR #941, opened 2026-07-31
feat(macos): native Connect Client form — preview-first, token-guarded (spec 091)

Dumbris:
## What

A native tray path for connecting AI clients (Claude Code, Cursor, Windsurf, VS Code, Codex, Gemini, OpenCode, Claude Desktop) — "Connect Client…" menu item + SwiftUI form replacing the Web-UI-only flow and the legacy preview-less dashboard sheet:

- **TCC-safe list**: stat-only aggregate (config present / no config / unsupported) — zero config-content reads until the user selects a client.
- **Preview before write, structurally**: the Connect control cannot exist before a preview renders — pending entry, config path, a whitelist-by-construction summary of the entry being replaced (names/types/query+userinfo-stripped endpoint, header/env *names* only — no secret value can appear), backup consequences, credential-embedding notice, and the core's refusal (e.g. OpenCode without a config) surfaced up front.
- **Drift-proof writes**: previews carry a keyed HMAC precondition token over the raw pre-write state (deterministic, single-resolution adoption lookup) *and* the pending entry; any drift — external edits (even ones the summary hides), file appearance/disappearance, proxy-side credential/address changes — yields a discriminated 409 and an automatic re-preview. No silent overwrites: `force` never rescues a stale token.
- **Session-scoped Undo** (restores the backup / removes a created file) and confirmed Disconnect.
- **Strict-socket mutations**: writes fail closed if the admin socket disappears mid-session, never silently ride TCP.

Backend deltas are additive and back-compatible (absent token = legacy behavior; Web UI unchanged): 3 new preview fields, optional `precondition_token` on the write, typed 409 discriminator, plus a SC-006 socket-to-gated-route e2e test.

## Spec & reviews

Full speckit artifacts in `specs/091-connect-client-form/` — spec approved after 4 Codex rounds; plan/tasks/implementation reviewed by an internal adversarial multi-agent panel with refute-verification (Codex quota-locked and Gemini CLI sunset mid-pipeline; 12 confirmed code findings all fixed + closure-verified, incl. a nondeterministic adoption-lookup class proven empirically).

## Testing

- Swift: 561 → 604 tests, 0 failures (state-machine invariants, decode fallbacks for pre-091 cores, in-flight/interleaving races, window-lifecycle teardown probed against real AppKit sheets, mutation-verified poll cancellation).
- Go: `internal/connect` + `internal/httpapi` green incl. no-raw-secrets suites (rotated keys, bearer/env secrets, `?apikey=`, `user:pass@`); full `./internal/...` green; linter 0 issues; `make swagger-verify` green.
- New docs: `docs/features/connect-clients.md` + refreshed REST reference.
- Outstanding (release-blocking): manual visual protocol (`specs/091-connect-client-form/verification/manual-protocol.md`) — NOT RUN, needs a live app run.

## Deliberate follow-ups

- Race-scale TOCTOU between token check and write (pre-existing class, shrunk from human-scale; full closure needs file locking).
- A legacy URL-equivalent duplicate entry is now never auto-deleted (correct trade: never delete what the preview never named); manual cleanup remains.
- macOS 13: `selectionDisabled` unavailable — keyboard skip of unsupported rows is 14+; model-side guard remains the backstop.


Dumbris:
## CodeQL check: pre-existing false positive (not introduced here)

The failing CodeQL run reports `go/path-injection` (high) at `internal/connect/connect.go` — **alert #164, created 2026-07-03 and still open on `main`**. This PR changes 168 lines in that file, so CodeQL's diff-informed analysis re-attributes the pre-existing alert to the diff; its own summary hedges: *"Alerts not introduced by this pull request might have been detected because the code changes were too large."*

**Verified it is genuinely unreachable, not merely old:** the `{client}` URL parameter never reaches a path join. `ConfigPath` switches on the closed set of 8 registry IDs and returns `""` for anything else, so traversal payloads produce no path at all. Proven empirically, and the proof is committed as a permanent regression guard (`internal/connect/path_injection_test.go`) — if the switch is ever replaced by string interpolation, that test fails and the alert stops being a false positive.

`CodeQL` is not among the 9 required status checks, so it does not gate merge. Same class as the previously-dismissed `go/request-forgery` config-URL FP in this repo.

Issue #936, opened 2026-07-30
tray: no way to point a dev/QA instance at an alternate root — every test run mutates real ~/.mcpproxy state


Dumbris:
Dev-ergonomics issue, not a user-facing bug. Surfaced repeatedly while doing live QA of #927 and #930, where it forced awkward workarounds and caused one accidental mutation of real user state.

## The problem

The tray resolves its paths via `homeDirectoryForCurrentUser`, which **ignores `$HOME`**. So a tray built from a branch and run out of a `/tmp` bundle copy still reads and writes the real `~/.mcpproxy`. There is no instance-root override.

Two concrete consequences observed:

**1. The socket path is fixed.** `~/.mcpproxy/mcpproxy.sock` is the only path a dev tray will use, so an isolated core must be started with `--tray-endpoint unix:///Users/<user>/.mcpproxy/mcpproxy.sock` to be found at all. It also makes the socket a machine-wide exclusive resource: two QA runs cannot proceed in parallel, and one will silently attach to the other's core and produce garbage results. That happened during this batch and had to be resolved by explicit sequencing.

**2. The autostart sidecar silently overwrites real user state.** `AutostartSidecarService.refresh()` reads `AutoStartService.isEnabled` (the *running* bundle's `SMAppService` login-item state) and writes it to `~/.mcpproxy/tray-autostart.json` on every app launch. A copied bundle under `/tmp` is not a registered login item, so it writes `{"enabled":false}` straight over the user's `{"enabled":true}`. Observed and restored by hand during QA; the value was wrong for about seven minutes. Impact while wrong is limited to the core's telemetry autostart reporting — the actual login-item registration is only ever read, never changed — but the write is silent and unexpected.

For a normal user with a single installed app, both behaviours are correct. The gap is that there is no supported way to run a second instance.

## Suggested direction

#926 already made `socketPath` injectable for tests. Extending that to the sidecar path would fix the immediate hazard, but the cleaner fix is a single **instance-root override** — an env var (e.g. `MCPPROXY_HOME`) or launch argument that relocates the socket, the sidecar file, and anything else resolved under `~/.mcpproxy` in one move.

That would make tray QA safe by construction rather than by discipline, and would let independent tray runs proceed in parallel instead of being serialized on a hardcoded path.

Dumbris:
## Partial workaround found: `CFFIXED_USER_HOME` does work

[...]

Issue #934, opened 2026-07-30
tray glance: error detail is truncated mid-identifier, and the header count can disagree with the list beneath it


Dumbris:
Two low-severity findings from live QA of the glance section (#930). Both are cosmetic; neither blocks that PR.

## 1. Error detail cut mid-identifier

`GlanceSection.firstClause` splits the error message on the first `.`, `:` or newline. Because mcpproxy error text routinely embeds a `server:tool` identifier, the split lands inside the identifier. Observed:

- Real core message: `invalid arguments for memory:create_entities: at '/entities': got string, want array`
- Rendered: `memory:create_entities · invalid arguments for memory — 8s`

It reads as though the server named `memory` were at fault. This is the common case for mcpproxy errors, not an edge case. The full message is still available in the tooltip.

Suggested fix: skip separators that fall inside a `server:tool` token, or split on `": "` (colon-space) rather than a bare colon.

## 2. Header count disagrees with the list

Two separate causes, same symptom:

- The header counts **all** clients while the list caps at 5 rows, with no "+N more" affordance — observed "11 clients" above 5 rows.
- The header count comes from the 30s usage poll while the rows arrive over SSE, so immediately after a call the header can contradict the rows below it — observed header "12 calls this hour" sitting above a 5-second-old row that had already made it 13.

Suggested fix: add a "+N more" row when truncating, and derive the header from the same SSE-updated state the rows use so the two cannot drift.

Dumbris:
Fixed in #945 (merged as `b30407fea`).

**(a) Clause truncation.** The error-detail splitter no longer cuts inside a `server:tool` identifier:
```
input : invalid arguments for memory:create_entities: at '/entities': got string, want array
  before: invalid arguments for memory                 <- accuses the server
  after : invalid arguments for memory:create_entities
input : failed to connect to 127.0.0.1:8080: connection refused
  before: failed to connect to 127     after: failed to connect to 127.0.0.1:8080
```
Sentence and newline splits are unchanged. Any `…` still seen in the menu is the display-width truncation applied *after* the clause is kept whole, not the splitter.

**(b) Header/list disagreement.** Verified reconciled live: 8 extra MCP sessions → menu showed 5 client rows + `+4 more` under a header reading `9 active` (5+4=9). Call count tracked correctly too: baseline `1 call this hour` → 3 real tool calls → `4 calls this hour` within 3s, stable across later polls (no double-count, no drop-back).

[...]

Issue #933, opened 2026-07-30
tray: a live core with a saturated listen backlog can be launched over, producing doomed spawns and a misleading error


Dumbris:
Found by cross-model review (round 6) of #927, which introduced the stale-socket escalation. Not a merge blocker for #927 — the single-writer invariant still holds — but the failure mode is user-visible and confusing.

## What happens

`onlyEverRefused` is used as evidence that a socket file is stale (refusals > 0 and nothing ever accepted during the episode). That is sound for a crashed core, but a **live** core whose listen backlog is saturated *before the episode begins* refuses every probe too. Since it never accepts once, the evidence never records `somethingAccepted`, and after the 60s deadline `reportUnresponsiveCore` escalates to a launch (`CoreProcessManager.swift:792`). The escalation's own confirmation and the launch preflight also only ever see refusals, so neither vetoes it.

## Consequence

The tray launches cores that are doomed: each new core blocks ~10s on bbolt's exclusive lock, fails, and the ladder retries. The attach watch has already been cancelled. The user eventually sees a retry-exhaustion / database-locked error about a core that is actually alive and healthy — just briefly saturated.

Data integrity is fine throughout: bbolt's exclusive flock is acquired before socket cleanup, so there is never a second writer.

## Why the existing test does not cover it

`CoreLivenessTests.swift:861` (`testACoreThatAcceptedEarlierInTheWindowIsNeverEscalatedOver`) models a core that accepts early and then refuses — but it stops the listener entirely after that acceptance, which makes the socket genuinely stale. The uncovered case is a listener that is **live for the whole episode** and refuses every probe from the start.

## Possible directions

- Treat a socket file whose inode/mtime is changing, or which a process still holds open, as evidence against staleness (`lsof`-equivalent via `proc_pidfdinfo`, or check the file is still bound).
- Have the spawned core distinguish "database locked" from "failed to start" and, on lock contention, tell the tray to go back to attaching rather than climbing the retry ladder.
- Cap escalation at one attempt per episode, so the worst case is one doomed spawn rather than a ladder.

Dumbris:
## Reproduced live, with two corrections to the severity assessment

[...]

Issue #862, opened 2026-07-16
macOS: silent tray/core exits are unattributable — add shutdown-reason logging + lifecycle diagnostics (was: phantom v0.50.0 self-relaunch report)


algis-dumbris:
## Summary

On v0.50.0 (Personal, macOS DMG install), the whole app — tray **and** core — silently exited and self-relaunched in the middle of an active working session, ~25h after install. There was no user action, no macOS sleep/reboot at that moment, no crash report, no version change, and **no shutdown reason logged**. The ~60s outage dropped 3 active MCP sessions; a connected streamable-HTTP client (Claude Code) marked the server disconnected and required a manual reconnect to recover.

Previous versions I ran (v0.41 → v0.48) never restarted spontaneously — every `Starting mcpproxy` line in my logs before v0.50.0 corresponds to a manual upgrade/restart (v0.48.0 ran Jul 9–13 continuously, surviving multiple macOS sleep cycles). This looks like a lifecycle regression in v0.50.0, possibly around the new "Start Core when the app opens" feature (#847, #410).

## Timeline (local time, +03:00)

- **Jul 14 12:16** — v0.50.0 installed and started (DMG). Runs normally for ~25h.
- **Jul 15 12:59:06** — last log line from the old core (routine `Doctor diagnostics completed`, running on its normal ~31s cadence all morning — machine awake, client session active).
- **Jul 15 12:59:06 → 13:00:04** — process dies **silently**: no shutdown/exit lines in main.log, no crash report in `~/Library/Logs/DiagnosticReports`, no reboot (uptime since Jun 15), binary mtime unchanged (no update applied).
- **Jul 15 13:00:04** — tray (ppid 1) and core both start fresh: `Starting mcpproxy {"version": "v0.50.0"}` … `Closed stale sessions on startup {"count": 3}`. `~/.mcpproxy/tray-autostart.json` is rewritten the same second (`"updated_at":"2026-07-15T10:00:04Z"`), so *something* orchestrated a relaunch — but nothing logged why.
- Connected MCP client (Claude Code, streamable HTTP on `127.0.0.1:8080/mcp`) hit the outage, flagged the server disconnected, and stayed broken until a manual `/mcp` reconnect.

## Log excerpts

Old instance, last minute (normal operation, then nothing):

```
2026-07-15T12:59:06.124+03:00 | INFO | httpapi/server.go:3302 | Doctor diagnostics completed | {"total_issues": 0, ...}
[...]

PR #945, opened 2026-08-01
fix(tray): glance clause/header consistency, saturated-core guard, instance root, lifecycle journal

Dumbris:
Four macOS tray issues, one branch. All Swift; no Go touched.

## #934 — glance error clause cut mid-identifier, header disagreeing with the list

**Error clause.** `firstClause` split on the first bare `.`, `:` or newline, which lands inside the two tokens mcpproxy errors are most likely to carry — a `server:tool` identifier and a host:port. `invalid arguments for memory:create_entities: at '/entities': got string, want array` rendered as `invalid arguments for memory`, reading as an accusation against the server. A separator is now a clause boundary only when whitespace or the end of the message follows it. Full message still in the tooltip.

**Header vs list**, both causes from the issue:

- The Clients segment counts every client while the list caps at five. A truncated list now carries a `+N more` row. Gaining or losing that row is structural (waits for menu close, FR-023); N alone is rewritten in place, so a twelfth client cannot leave `+6 more` under a header that counted twelve.
- The call count came from the 30s usage poll while rows arrive over SSE. The header now reads `AppState.glanceCallsThisHour(now:)` — the polled total plus live calls the poll has not answered for, dropped the moment it does, so the two cannot accumulate. Only records the core's own aggregate counts (`type == "tool_call"`, per `UsageAggregate.Apply`) are added, so the correction cannot overshoot and jump back at the next poll.

## #933 — a saturated-but-live core is no longer launched over

`onlyEverRefused` read "refusals, nothing ever accepted" as proof of a stale socket. On macOS a full listen backlog is `ECONNREFUSED` too, so a core already saturated when the episode began is indistinguishable from a dead core's leftovers to every socket probe there is.

New signal, independent of the socket: `DataDirectoryLock` asks whether a live process holds bbolt's exclusive `flock` on the data directory (`<socket dir>/config.db` — derived from the socket, so a tray pointed at another instance asks about that instance). The kernel releases `flock` on process death, so unlike a leftover socket file it cannot be stale. It is taken with `LOCK_SH | LOCK_NB` and released immediately: shared still conflicts with bbolt's exclusive lock, while two trays probing at once do not report each other as live cores.

When the lock is held the tray re-arms the deadline and keeps waiting, with the attach watch still running underneath, so the core is picked up as soon as it drains — instead of spawning cores that die on that same lock and ending on a retry-exhaustion error about a healthy core. A refusing socket whose database nobody holds is still escalated over, so a genuinely stale socket does not deadlock startup.

## #936 — instance-root override (the half that had not shipped)

`InstancePaths` resolves the socket, config, database, autostart sidecar and lifecycle journal in one place, relocatable with `MCPPROXY_HOME`. A core this tray spawns is handed the same root (`serve --data-dir <root> --config <root>/mcp_config.json`); with the variable unset the command is byte-for-byte `["serve"]` as before and every path resolves exactly where it did. `MCPPROXY_SOCKET_PATH` still outranks the root (it names one core, not one instance). A socket path over `sun_path`'s 103 bytes is reported at launch rather than failing to bind in silence. Documented in `docs/tray-debug.md`, including what is deliberately NOT relocated (the core's log directory, and the preferences domain — `cfprefsd` honours neither `$HOME` nor this variable).

## #862 — shutdown-reason logging and lifecycle diagnostics

`<instance root>/tray-lifecycle.jsonl` (JSON lines, so a kill mid-write costs the last record and not the file) plus `os_log` at **Notice** — the Info tier is purged within hours, which is why the original incident had no context left. Records: `appLaunched`, `appTerminating`, `signalReceived`, `coreLaunched`, `coreTerminated`, `coreExited`, `updateCheck`; each with a reason, the tray's uptime and a pid.

- The reason is claimed in advance by whoever initiates the stop (first claim wins, so "user chose Quit" is not overwritten by the core teardown it causes). `applicationWillTerminate` is told the app is going, never why.
- A shutdown nobody claimed is recorded as `unattributed (no initiator claimed this shutdown)` rather than as something plausible.
- `SIGTERM`/`SIGINT`/`SIGHUP` are caught, recorded and routed through the normal quit path; logout/restart is claimed via `NSWorkspace.willPowerOffNotification`.
- A run whose last record is not an `appTerminating` is reported at the NEXT launch, with the last thing it did and how long it had been up — the only way a SIGKILL-class death (jetsam, `pkill`) is attributable at all.
- One line per periodic update check (ask 3).

Under XCTest the shared journal goes to a scratch file: the suite drives the real `CoreProcessManager`, so without that it would append phantom cores to the developer's own live install. Pinned by a test.

## Verification

`cd native/macos/MCPProxy && swift test`

```
Test Suite 'All tests' passed at 2026-08-01 08:29:38.
	 Executed 766 tests, with 0 failures (0 unexpected) in 45.338 (45.384) seconds
```

Baseline before this branch was 720 tests; 46 added.

Red-before-green, e.g. #933. With `DataDirectoryLock.probe` stubbed to always answer `.free` (i.e. the pre-fix behaviour), the new test reproduces the reported bug exactly — doomed spawns over a live core:

```
SaturatedCoreEscalationTests.swift:71: error: -[MCPProxyTests.SaturatedCoreEscalationTests
  testASaturatedButLiveCoreIsNeverLaunchedOver] : XCTAssertEqual failed: ("2") is not equal to ("0")
  - a live process holds the data directory: that core is there, whatever its listen queue is doing
	 Executed 2 tests, with 3 failures (0 unexpected) in 6.388 seconds
```

and with the real probe:

```
	 Executed 2 tests, with 0 failures (0 unexpected) in 6.148 seconds
```

The #934 and #936/#862 suites were likewise written first and run red (missing symbols / wrong header count) before the implementations landed.

No Go files touched, so `scripts/run-linter.sh` and `make swagger` are not in play. Nothing under `~/.mcpproxy` was created or modified by the test runs (checked explicitly after each full run — the journal isolation test exists because the first run did create one).

## Deliberately deferred

- **#933, the terminal state.** The issue's follow-up comment notes that after the retry ladder exhausts, the tray stays wedged on "Maximum reconnection attempts exceeded" even once a core answers again, and the hint points at a Retry menu item that does not exist in that state. That predates #927 and is a separate fix to the ladder's end state; this PR removes the path that made it reachable from a healthy core.
- **#862, core-side items.** Periodic RSS watermark logging and a single-instance guard are core (Go) work; the core already logs its own signal and shutdown reason (`cmd/mcpproxy/main.go`), so the gap this PR closes is the tray side. Streamable-HTTP session resilience across restarts is explicitly tracked separately.
- **#936, preferences domain.** `cfprefsd` ignores both `$HOME` and `MCPPROXY_HOME`; a dev instance still needs a distinct bundle id for isolated defaults, and the first-run gate lives there. Documented rather than worked around.

### DIFF
--- a/cmd/mcpproxy/main.go
+++ b/cmd/mcpproxy/main.go
@@ -714,6 +714,15 @@ func loadConfig(cmd *cobra.Command) (*config.Config, error) {
 
 	// Load configuration - use LoadFromFile if config file specified, otherwise use Load
 	if configFile != "" {
+		// A named config that has never been written is a FIRST RUN, not an
+		// error: the relocated tray instance of GH #936 spawns its core with
+		// --data-dir <root> --config <root>/mcp_config.json against a root
+		// nothing has used before. The implicit path has always created its
+		// default here; the explicit one used to exit with "no such file or
+		// directory" instead, which made that whole flow unusable.
+		if _, ensureErr := config.EnsureConfigFile(configFile, dataDir); ensureErr != nil {
+			return nil, ensureErr
+		}
 		cfg, err = config.LoadFromFile(configFile)
 	} else {
 		cfg, err = config.Load()
--- a/docs/tray-debug.md
+++ b/docs/tray-debug.md
@@ -38,6 +38,8 @@ With it set:
 
 Notes and limits:
 
+- A brand-new root needs nothing prepared. The core creates the config file named by `--config` when it does not exist yet (printing `INFO: Created default configuration file at …`), so `export MCPPROXY_HOME=/tmp/mcpproxy-qa` and launching the tray is the whole setup. An existing file is never touched.
+- The core also reads its autostart sidecar out of the data directory it was given, so a QA instance reports its own login-item state rather than the real install's.
 - `MCPPROXY_SOCKET_PATH` still wins over the root. Use it to attach to a core somebody else started; use `MCPPROXY_HOME` to own a whole instance.
 - **Not** relocated: `~/Library/Logs/mcpproxy` (the core's log directory, which follows the core's own rules) and the app's preferences domain — `cfprefsd` honours neither `$HOME` nor this variable, so give a dev bundle a distinct bundle id if you need separate defaults.
 - The first-run dialog is a `UserDefaults` flag, so it lives in the preferences domain rather than the instance root; a fresh bundle id will show it again. Its "Launch at login" checkbox is pre-checked, so clicking through it in a dev instance registers a real login item.
@@ -55,8 +57,8 @@ Recorded events: `appLaunched`, `appTerminating`, `signalReceived`, `coreLaunche
 The rules worth knowing when reading one:
 
 - Every shutdown records **who asked**. A shutdown nobody claimed is written as `unattributed (no initiator claimed this shutdown)` rather than as something plausible.
-- `SIGTERM`, `SIGINT` and `SIGHUP` are caught, recorded, and then routed through the normal quit path. `SIGKILL` and a jetsam kill cannot be caught by anyone.
-- Which is why the **next** launch reports an unaccounted-for previous run: if the last record before a launch is not an `appTerminating`, the new `appLaunched` record says so and names the last thing the dead run did. That marker is the signature of a SIGKILL-class death (jetsam, `pkill`, power loss) or a crash outside the app's handlers.
+- `SIGTERM`, `SIGINT` and `SIGHUP` are caught, recorded, and then routed through the normal quit path — handled off the main thread, so a wedged main thread cannot swallow them. If the app has not gone five seconds later, the signal's default action is restored and re-raised: catching a termination signal delays the exit, it never prevents it. `SIGKILL` and a jetsam kill cannot be caught by anyone.
+- Which is why the **next** launch reports an unaccounted-for previous run: if the previous run recorded no `appTerminating` at all, the new `appLaunched` record says so and names the last thing the dead run did. That marker is the signature of a SIGKILL-class death (jetsam, `pkill`, power loss) or a crash outside the app's handlers. It is not the *last* record that decides this — an ordinary clean exit writes `appTerminating` first and then `coreTerminated` for the core it tears down.
 - The journal is trimmed to its newest 500 records at launch.
 
 ## Use Cases
--- a/docs/tray-debug.md
+++ b/docs/tray-debug.md
@@ -6,9 +6,58 @@ This guide explains how to control the mcpproxy tray during development and auto
 
 | Variable | Scope | Default | Purpose |
 |----------|-------|---------|---------|
+| `MCPPROXY_HOME` | macOS tray | `~/.mcpproxy` | Relocates the whole instance root: socket, config, database, autostart sidecar, lifecycle journal. |
+| `MCPPROXY_SOCKET_PATH` | macOS tray | `<root>/mcpproxy.sock` | Points the tray at one specific core's socket, wherever the root is. |
 | `MCPPROXY_TRAY_SKIP_CORE` | Tray | unset | Prevents the tray from launching the core binary. |
 | `MCPPROXY_CORE_URL` | Tray | `http://localhost:8080` | Overrides the core API endpoint the tray connects to. |
 | `MCPPROXY_DISABLE_OAUTH` | Core | unset | Disables OAuth popups and tray-driven login prompts. |
+
+## Running a Second (Dev/QA) macOS Tray Instance
+
+The native macOS tray resolves its paths through `homeDirectoryForCurrentUser`, which **ignores `$HOME`**. Without an override, a tray built from a branch and run out of a scratch bundle still reads and writes the real `~/.mcpproxy` — which means QA runs cannot go in parallel (they contend for one socket) and the autostart sidecar overwrites the user's real login-item state on every launch.
+
+`MCPPROXY_HOME` moves the whole instance in one step. It is unset in normal use, and with it unset every path resolves exactly where it always did.
+
+```bash
+# Keep the root SHORT: sockaddr_un caps the socket path at 103 bytes, and a
+# deep scratch directory silently fails to bind. The tray logs a warning if
+# the resolved path is over the limit.
+export MCPPROXY_HOME=/tmp/mcpproxy-qa
+/path/to/dev/mcpproxy.app/Contents/MacOS/MCPProxy
+```
+
+With it set:
+
+| File | Default | With `MCPPROXY_HOME=/tmp/mcpproxy-qa` |
+|------|---------|----------------------------------------|
+| Core socket | `~/.mcpproxy/mcpproxy.sock` | `/tmp/mcpproxy-qa/mcpproxy.sock` |
+| Config opened by **Open Config File** | `~/.mcpproxy/mcp_config.json` | `/tmp/mcpproxy-qa/mcp_config.json` |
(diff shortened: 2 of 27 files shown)


Dumbris:
## Cross-model review round — all 7 findings fixed (TDD)

Four commits on top of `cf022b6a2`. Every fix had its failing test written and run first; red output quoted per finding.

### P1

**1. A fresh `MCPPROXY_HOME` instance could not start at all** — `ba28f7416`

The tray emits `--config <root>/mcp_config.json`, and `LoadFromFile` fatals on a missing explicit path, so the documented flow died on `no such file or directory`. Fixed at the **serve entry point**, not inside `LoadFromFile`: that function is also the hot-reload path (`internal/runtime/lifecycle.go`, `internal/runtime/configsvc`), where a config file that vanishes under a running proxy must fail the reload and keep the live config — never silently become a default with no servers. New `config.EnsureConfigFile(path, dataDir)` seeds one only when nothing is there, prints the same `INFO: Created default configuration file at …` notice the implicit path prints, and never touches an existing file. `TestLoadFromFileStillRefusesAMissingExplicitPath` pins that the hot-reload path stayed strict.

Red:
```
internal/config/ensure_config_file_test.go:23:18: undefined: EnsureConfigFile
FAIL	github.com/smart-mcp-proxy/mcpproxy-go/internal/config [build failed]
```
Green (4 tests) + **live**, using the exact argv the tray emits against a root that did not exist, on an isolated port:
```
$ rm -rf /tmp/mcpproxy-qa-wf8
$ mcpproxy serve --data-dir /tmp/mcpproxy-qa-wf8 \
      --config /tmp/mcpproxy-qa-wf8/mcp_config.json --listen 127.0.0.1:18142
INFO: Created default configuration file at /tmp/mcpproxy-qa-wf8/mcp_config.json
… "Configuration loaded" {"data_dir": "/tmp/mcpproxy-qa-wf8", "servers_count": 0}
$ curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:18142/ready
200
```

**2. Once the data-directory lock was seen held, the tray could wait forever** — `77d01e42b`

Confirmed: `dataDirectoryHeld` latched, and `onlyEverRefused` — the only gate that re-probed — was made permanently false by that same flag. A saturated core that then died (SIGKILL/jetsam) left its socket behind and every later deadline re-armed on a fact that had stopped being true, in a state offering neither Stop nor Retry.

A lock is a fact about *now*, not accumulated evidence (the kernel drops `flock` when the holder dies), so it is asked fresh at each deadline and never remembered between them; `onlyEverRefused` goes back to being purely what the socket said. New test releases the lock mid-episode and requires escalation.

Red:
```
SaturatedCoreEscalationTests.swift:149: error: -[… testACoreThatDiesAfterTheFirstDeadlineIsEventuallyEscalatedOver] :
  XCTAssertGreaterThan failed: ("0") is not greater than ("0") — nothing holds the data directory any more
```
Green — and the two existing cases (live saturated core is never launched over; a genuinely stale socket still is) both still pass.

**3. Every logout / signal shutdown reported as a SIGKILL-class crash** — `c6af016ad`

Confirmed. `applicationWillTerminate` writes `appTerminating` and then `coreTerminated`, and the classifier read only the last record. The question is whether the previous **run** recorded a reason, so it is now asked of the records since that run's `appLaunched`. Two tests: the production ordering is clean, and an earlier run's goodbye does not vouch for the run after it.

Red:
```
AppLifecycleTests.swift:149: error: … XCTAssertNil failed: "previous run ended without recording a
  shutdown reason — last event coreTerminated "tray is terminating" … this was SIGKILL-class"
LifecycleJournalTests.swift:117: error: … got unclean(last: … coreTerminated …)
```

### P2

**4. Header could jump backwards, or overcount, around each usage poll** — `0d16dc4f6`

Confirmed. Fixed with a single boundary: the instant the poll was **issued** (stamped before the `await` in `refreshUsage`). Live calls strictly newer than it survive the response; calls at or before it are assumed already counted and are neither re-admitted nor added twice. Two tests, one per direction (in-flight poll must not erase a call it could not have seen; a late SSE event for an already-counted call must not be added).

**5. `glanceCallsThisHour` filtered live calls by hour but not the polled base** — `0d16dc4f6`

Confirmed, including that the old test enshrined the wrong value. `callsThisHourBucket` records which UTC hour the polled total is *for*; after a rollover it is dropped rather than carried. The old `…IsNotCountedInThisOne` assertion of `12` is replaced by `testNeitherHalfOfTheCountSurvivesAnHourRollover` (expects `0`) plus `testAfterAnHourRolloverOnlyTheNewHoursLiveCallsCount`.

Red:
```
GlanceHeaderConsistencyTests.swift:158: error: … XCTAssertEqual failed: ("Optional(12)") is not equal to ("Optional(0)")
GlanceHeaderConsistencyTests.swift:173: error: … XCTAssertEqual failed: ("Optional(13)") is not equal to ("Optional(1)")
```

**6. `SIG_IGN` on TERM/INT/HUP with main-queue-only delivery** — `c6af016ad`

Both halves confirmed and both fixed. `SIG_IGN` → a no-op `sigaction` handler: it suppresses the default action just as well, is all `EVFILT_SIGNAL` needs, and is **reset across `execve`**, so spawned cores no longer inherit an ignored SIGINT/SIGHUP. Delivery moved off `.main` onto its own queue (the terminate closure makes the hop to main itself), and if the app is still here 5 s later the signal's default action is restored and re-raised — catching a termination signal may delay the exit, never prevent it.

Three tests, including one that blocks the main thread on a semaphore and requires a real `SIGHUP` to still be acted on. Red:
```
AppLifecycleTests.swift:196: error: value of type 'AppLifecycle' has no member 'uninstallSignalHandlers'
AppLifecycleTests.swift:218: error: extra arguments at positions #2, #3 in call
AppLifecycleTests.swift:238: error: value of type 'AppLifecycle' has no member 'respondToSignal'
```

**7. The relocated autostart sidecar was written but never read** — `ba28f7416`

Confirmed. `AutostartReaderForDataDir(dataDir)` follows the data directory; both call sites (`/api/v1/status`, the heartbeat's lazy init) pass the running config's `DataDir`, and an empty one keeps the historical `~/.mcpproxy`. Live, against a QA instance while the real install's sidecar said the opposite:
```
# no sidecar in the QA root
autostart_enabled = None
# after writing {"enabled":true} into /tmp/mcpproxy-qa-wf8
autostart_enabled = True   (real ~/.mcpproxy/tray-autostart.json says enabled:false)
```

### Gates

```
./scripts/run-linter.sh          → 0 issues
swift test                       → Executed 777 tests, with 0 failures (0 unexpected) in 50.041s
go test ./internal/config/ ./internal/telemetry/ ./internal/httpapi/ ./cmd/mcpproxy/ -count=1
  ok internal/config 4.290s | ok internal/telemetry 7.868s | ok internal/httpapi 0.581s | ok cmd/mcpproxy 0.860s
```
Pre-push hooks ran `Verify OpenAPI spec is up to date` (passed) — no REST annotations were changed, only a handler body.

`docs/tray-debug.md` updated on three points the fixes made stale: a fresh root needs no preparation, the sidecar follows the data directory, and the unclean-exit marker is not decided by the last record.

### Deliberately not done

- `EnsureConfigFile` will also create a config for a **typo'd** `--config` path rather than erroring. That is the same behaviour the implicit path has always had, and it prints the same `INFO: Created default configuration file at …` line, so it is visible rather than silent. Making it stricter would require distinguishing "first run" from "typo", which nothing on the command line can do.
- The poll boundary uses the request-issue time, so a call landing inside one HTTP round trip of the snapshot can still be counted twice for up to one poll interval. Closing that fully needs call identity in the usage aggregate, which the endpoint does not return — the remaining window is milliseconds wide, against the 30-second one it replaces.

Issue #922, opened 2026-07-28
connect opencode fails: OpenCode bootstraps opencode.jsonc, mcpproxy only recognizes opencode.json


algis-dumbris:
## Summary

`mcpproxy connect opencode` (and the Connect Clients flow) fails on a fresh OpenCode installation because recent OpenCode versions bootstrap their global config as **`opencode.jsonc`**, while mcpproxy only recognizes **`opencode.json`**. OpenCode is then reported as not installed, and an explicit connect is refused.

## Environment

- MCPProxy v0.52.0 (personal) darwin/arm64 (tray app + bundled CLI)
- OpenCode 1.18.8 (Homebrew), macOS
- Fresh OpenCode setup: first run auto-created `~/.config/opencode/` with `opencode.jsonc` containing only the `$schema` stub — no `opencode.json` anywhere

## Reproduction

```
$ opencode --version
1.18.8
$ ls ~/.config/opencode/
.gitignore  node_modules  opencode.jsonc  package-lock.json  package.json

$ mcpproxy connect opencode
Error: OpenCode config file /Users/user/.config/opencode/opencode.json does not exist
```

The status/UI side has the same blind spot: OpenCode shows `Exists: false` (treated as not installed), so the Connect Clients flow offers no working path either.

## Root cause

`internal/connect/clients.go` — `ConfigPath("opencode", …)` returns only the `.json` variant:

```go
case "opencode":
    ...
    return filepath.Join(homeDir, ".config", "opencode", "opencode.json")
```

`internal/connect/connect.go` — `Connect()` has an opencode-specific existence gate on that single path, so a `.jsonc`-only installation can never connect:

```go
if client.ID == "opencode" {
    if _, err := os.Stat(cfgPath); os.IsNotExist(err) {
        return nil, fmt.Errorf("OpenCode config file %s does not exist", cfgPath)
    }
}
```

`GetAllStatus()` / `GetStatus()` use `os.Stat` on the same single path, so detection reports OpenCode as absent.

Meanwhile OpenCode itself probes and merges **three** global config files (from its own log, v1.18.8):

```
loading path=~/.config/opencode/config.json
loading path=~/.config/opencode/opencode.json
loading path=~/.config/opencode/opencode.jsonc
```

and its bootstrap creates `opencode.jsonc`, not `opencode.json`. So on any freshly installed OpenCode, mcpproxy's single-path assumption never matches.

## Suggested fix

[...]

PR #949, opened 2026-08-03
fix(connect): name the checked config paths on "No config found" rows

Dumbris:
## Problem

On a laptop running stable mcpproxy (≤ v0.52.1), the Connect Client form showed **"No config found"** for OpenCode even though `~/.config/opencode/opencode.jsonc` exists. Two separate issues:

1. The `.jsonc` support (#922, fixed by #923) has only shipped in `v0.53.0-rc.*` prereleases — stable builds only check `opencode.json`. Upgrading resolves the detection itself.
2. The form never says **which paths it checked**, so a mismatch like this is undiagnosable from the UI. That is what this PR fixes.

## Changes

- **Core**: `ClientStatus` gains an additive `checked_paths` field — every config location the existence check consults, highest precedence first. For OpenCode that is both `opencode.jsonc` and `opencode.json`; for other clients the single `config_path`. Populated on both the stat-only aggregate (`GET /api/v1/connect`) and the per-client detail. Static path list, no extra stats/reads — the content-read-free contract (Spec 075) is untouched.
- **macOS form**: supported, unconnected rows without a config now render a `Looked for opencode.jsonc or opencode.json in ~/.config/opencode` note (home-abbreviated). A denied-access remediation or an explicit core note still outranks it. Against a core older than the app, it falls back to naming `config_path`.
- **Web UI**: already shows `config_path` per row; only the hand-written `ClientStatus` type gained the new optional field.

## Tests

- Go: `checked_paths_test.go` — aggregate + detail carry the candidates, `.jsonc` first; single-file clients report exactly one path. `go test -race ./internal/connect/...`, strict golangci-lint v2, and `./scripts/test-api-e2e.sh` all pass.
- Swift: fixture-driven model tests for the looked-for note (multi-candidate collapse, single-path fallback, no note when config present, core note precedence). Full `swift test` suite passes.

### DIFF
--- a/frontend/src/types/api.ts
+++ b/frontend/src/types/api.ts
@@ -883,6 +883,9 @@ export interface ClientStatus {
   // privacy fix including the exact tccutil reset command).
   access_state?: AccessState
   remediation?: string
+  // Every config location the existence check consults, highest precedence
+  // first (e.g. OpenCode's opencode.jsonc then opencode.json).
+  checked_paths?: string[]
 }
 
 export interface ConnectResult {
--- a/internal/connect/checked_paths_test.go
+++ b/internal/connect/checked_paths_test.go
@@ -1,0 +1,76 @@
+package connect
+
+import (
+	"path/filepath"
+	"runtime"
+	"testing"
+)
+
+// A "No config found" row must be able to name the exact files that were
+// looked for — otherwise a user whose opencode.jsonc exists cannot tell
+// whether mcpproxy checked the wrong file or could not see it.
+
+func statusFor(t *testing.T, statuses []ClientStatus, id string) ClientStatus {
+	t.Helper()
+	for _, st := range statuses {
+		if st.ID == id {
+			return st
+		}
+	}
+	t.Fatalf("client %q not in status list", id)
+	return ClientStatus{}
+}
+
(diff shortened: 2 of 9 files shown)


Dumbris:
Cross-model + adversarial review round 1 (codex quota-blocked; used opencode/gpt-5.6-sol + refute-verify panel) surfaced four genuine defects, fixed in f1f2dc31a:

- **Production bug, pre-existing since #923**: every production `connect.Service` is constructed without a `homeDir` (`NewService`), and `opencodeConfigCandidates("")` joined CWD-relative paths — so the `.jsonc` preference never actually fired in production builds (it only worked in tests, which use `NewServiceWithHome`). Empty home now resolves via `os.UserHomeDir`, fixing the jsonc preference, the disconnect/undo alternate-candidate scans, and the new `checked_paths` alike.
- A permission-blocked aggregate row (`access_state=denied`, which the stat-only list reports **without** remediation) fell through to the looked-for note, claiming files were checked when the stat was forbidden. The gate now only fires for absent/unknown.
- The looked-for hint rendered in the orange warning channel on every not-installed row and was invisible to VoiceOver; it now renders secondary and is included in the row accessibility label.
- The tilde-abbreviation branch had zero test coverage (fixtures used `/Users/x`); real-home tests added.

PR #950, opened 2026-08-03
feat(tray): render the 24h activity histogram inline at the top of the menu

Dumbris:
## What

The 24h activity chart used to hide behind an "Activity (24h)" submenu — the overview was the only glance row that required a second navigation step. It now renders **inline, directly under the summary line**, so the day's shape is on screen the moment the tray icon is clicked.

Verified live on a dev build (screenshot in the session): summary → chart → Recent → Clients, no submenu.

## How

- `GlanceSection` builds the row eagerly but **caches the chart item by its shaped bars** (`HistogramBar` is `Equatable`), so the debounced closed-menu rebuilds never re-render a SwiftUI chart nobody is looking at. `rebuildMenu` reuses one `NSMenu` via `removeAllItems()`, so re-adding the cached item is safe.
- The row's **kind** (loading placeholder / failure row / 132 pt chart) is structural in `updateInPlace`: a kind flip resizes the menu, so it defers to `menuDidClose` like every other structural change (FR-023 discipline). Within a kind, new bars swap the hosted view in place (fixed `chartItemSize` frame, no resize) and a failure row refreshes its tooltip.
- The submenu's "Not updating" marker is not ported: the inline row sits directly under the summary line, whose "not updating" segment already says it.
- `HistogramSubmenuDelegate` is deleted. The pure loading/failed/chart decision (`ActivityHistogram.state`) and the injectable chart factory seam are unchanged.
- Spec 048 invariant intact: building the row reads `AppState` only, no network request.

## Tests

Old submenu-structure tests are replaced by `GlanceInlineHistogramTests`: the three row kinds, chart-wins-over-stale-failure, timeline-arrives-while-closed, the eager-build cache (factory called once for unchanged bars, same item instance), kind-flip-is-structural both directions, in-place view swap under an open menu, tooltip refresh, and the real-chart size/accessibility checks. Full `swift test` suite: 725+ tests, 0 failures.

### DIFF
--- a/native/macos/MCPProxy/MCPProxy/MCPProxyApp.swift
+++ b/native/macos/MCPProxy/MCPProxy/MCPProxyApp.swift
@@ -898,12 +898,18 @@ final class AppController: NSObject, NSApplicationDelegate, NSWindowDelegate, NS
             menu.addItem(row)
         }
 
-        // Needs Attention — only auth required, connection errors, quarantine (NOT disabled)
+        // Needs Attention — only auth required, connection errors, quarantine
+        // (NOT disabled). One collapsed row: the count is the glanceable fact,
+        // the per-server detail is a hover away, and N servers no longer cost
+        // N rows of a menu that opens with a chart. Absent entirely when
+        // nothing needs attention.
         let attentionServers = appState.serversNeedingAttention
         if !attentionServers.isEmpty {
-            let header = NSMenuItem(title: "Needs Attention (\(attentionServers.count))", action: nil, keyEquivalent: "")
-            header.isEnabled = false
-            menu.addItem(header)
+            let parent = NSMenuItem(title: "Needs Attention (\(attentionServers.count))",
+                                    action: nil, keyEquivalent: "")
+            parent.image = NSImage(systemSymbolName: "exclamationmark.triangle",
+                                   accessibilityDescription: "needs attention")
+            let submenu = NSMenu(title: "Needs Attention")
 
             for server in attentionServers {
                 let action = server.health?.action ?? ""
@@ -924,8 +930,10 @@ final class AppController: NSObject, NSApplicationDelegate, NSWindowDelegate, NS
                 // by VoiceOver (same FR-025 discipline as the glance rows).
                 item.setAccessibilityLabel(fullTitle)
                 item.image = NSImage(systemSymbolName: icon, accessibilityDescription: action)
-                menu.addItem(item)
+                submenu.addItem(item)
             }
+            parent.submenu = submenu
+            menu.addItem(parent)
             menu.addItem(.separator())
         }
 
--- a/native/macos/MCPProxy/MCPProxy/MCPProxyApp.swift
+++ b/native/macos/MCPProxy/MCPProxy/MCPProxyApp.swift
@@ -331,11 +331,6 @@ final class AppController: NSObject, NSApplicationDelegate, NSWindowDelegate, NS
         // a full rebuild (removeAllItems) on a menu already on screen and
         // re-arm/disarm the guard under the parent — exactly the
         // restructuring-while-open the design forbids.
-        //
-        // The glance histogram submenu does build its single row lazily on open,
-        // but it does NOT reach this method: GlanceSection installs its own
-        // HistogramSubmenuDelegate on that submenu, precisely so opening it
-        // cannot rebuild the tray menu underneath the cursor.
         guard menu === menuHost?.menu else { return }
 
         // Spec 048: dropped the per-click `client.servers()` fetch. appState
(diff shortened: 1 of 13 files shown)


Dumbris:
Adversarial review round 1 (two independent lenses + opencode/gpt-5.6-sol cross-model; codex quota-blocked) — findings, all addressed in c3cb0edbf:

- **Real regression caught**: making the histogram row's kind structural froze the ENTIRE glance block (summary, row identities/click payloads, client presence) for the whole menu session when the timeline loaded while the menu was open — the common open-just-after-launch case. Main pins this behavior in `testTheTimelineArrivingWhileTheMenuIsOpenKeepsRowsUpdating`, which was red on the branch. Fixed: the histogram never vetoes the in-place update; text↔text rewrites in place, text↔chart keeps the built row until the next rebuild (`menuWillOpen` runs one before every display).
- A fetch failing while the menu is open now replaces "loading…" in place instead of leaving it up.
- The chart cache is additionally keyed by the current time zone (bars are UTC-keyed and survive a zone change; the VoiceOver "busiest hour" label does not).

Verified clean by the lifecycle lens on the current tip: cached-item re-add across rebuilds (single NSMenu, `removeAllItems()` on every path), hidden-block cache staleness, mid-open fixed-frame view swaps, no retain cycles, spec-048 no-fetch invariant, and structural discipline (no half-written menu path).

**Deliberate accessibility decision**: the old enabled "Activity (24h)" submenu anchor was keyboard-highlightable; the inline chart is a disabled view-backed item, skipped by arrow-key navigation (the chart data itself was equally keyboard-unreachable in the submenu — its only row was disabled). The full series summary stays on the host view for the VoiceOver cursor, and the summary text line above carries the headline numbers.

Note: `AppLifecycleTests.testTheSharedJournalNeverWritesToTheRealInstanceRootUnderTests` fails identically on main when a real tray run has written `~/.mcpproxy/tray-lifecycle.jsonl` — environmental, not this PR.

Dumbris:
Round 2 — compact revision (owner feedback: menu too wide, chart too big; screenshot-verified live):

- **Width discipline**: the widest thing in the menu is now the 24h chart block (272 pt), never text. Failed rows are one fact per line — title `label ×N — age`, error clause on the muted second line, displaced reason in the tooltip + VoiceOver string. Budgets: label 34→30, second line 60→44, pre-14.4 fallback clause 40→28. Needs-Attention rows truncate to the same budget with the full text in the tooltip.
- **Chart bounded**: 248×84 incl. legend (item 272×96, was 288×132) — a shape to recognise at a glance; the Web UI keeps the full-size chart.
- **Structural consistency**: a row *gaining* its error line defers to menu close, exactly like a run gaining its reason (FR-023); the identity-turnover/late-failure liveness tests pin the one-line fallback path where the rewrite lands in place, plus a new test pinning the structural case.

This is a deliberate compact revision of spec 090 FR-011a ("error joins the title") at the owner's request. Full suite: 787 tests, only the known environmental AppLifecycleTests failure (fails on main too).

Dumbris:
Round 3 (owner feedback): **Needs Attention collapses into a submenu** — one row with count + warning icon, per-server detail on hover, absent entirely when nothing needs attention. **An idle 24h axis renders words, not an empty chart**: muted "No calls in the last 24h" (new third text kind — swaps in place with loading/failed, text↔chart flips defer to the next rebuild like before).

Both screenshot-verified live on a dev build. Cross-model review (opencode) on the delta: no genuine defects — histogram state/cache transitions consistent, submenu ownership survives rebuilds, collapsed parent stays accessible. Suite: 787 tests, only the known environmental AppLifecycleTests failure.

Dumbris:
Round 4 (owner feedback): **one 24h frame for every number in the menu.** The reported contradiction — Recent rows from days ago under "No calls in the last 24h" — is gone:

- Header: "N calls in the last 24h" (`glanceCallsLast24h`, same poll+live-SSE reconciliation as GH #934, summed over the histogram axis; zero-count segment omitted — the idle row says it), summary row absent when it has nothing to say.
- Recent: filtered to the **hour-aligned** 24h axis; hidden entirely when nothing qualifies (Open Activity… stays).
- Clients placeholder: "No clients in the last 24h" (presence lookback was already 24h).

**Codex is back** and served as critic: round 1 caught a real edge (Recent used a raw 86,400 s cutoff vs the hour-aligned axis — a call 23 h 40 m old could show in Recent after its hour slid off the chart); fixed in b4be59ba9 with an edge test. Round 2: CLEAN, and it confirmed the poll/SSE snapshot watermark race it flagged is the pre-existing, deliberate GH-#934 trade-off on main, not worsened here.

Screenshot-verified live. Suite: 789 tests, only the known environmental AppLifecycleTests failure.

Issue #955, opened 2026-08-05
[Feature]: request queueing / concurrency limit for multi-user deployments


zeeid:
### Problem / motivation

We run mcpproxy as a shared HTTP service (systemd, headless mode) accessed by multiple concurrent users/agents. mcp_config.json currently has no way to cap concurrent requests or queue excess ones - bursts of simultaneous tool calls go straight to upstream servers, including stateful ones (e.g. MySQL-backed MCP servers), risking overload on those upstream resources when usage scales beyond a single user.

### Proposed solution

Add a config option (e.g. in mcp_config.json) to set:
- max_concurrent_requests (global and/or per-upstream-server)
- queue_size / queue_timeout for requests beyond the concurrency limit, so excess requests wait in a queue instead of failing or overwhelming the upstream

This would let a single mcpproxy instance safely serve multiple concurrent users without needing external infrastructure for basic overload protection.

### Alternatives considered

Currently working around this with an nginx reverse proxy (limit_req/limit_conn) in front of mcpproxy, and separately capping max_user_connections at the database level for MySQL-backed MCP servers. This works for HTTP-facing setups, but native support would help stdio-based upstream servers too, where reverse-proxy-level limits don't apply.

### Which edition?

Both / not sure

### Pre-submission checklist

- [x] I searched existing issues and discussions for a similar request

Dumbris:
Thanks — good timing, we just finished a design pass on this. Plan is native limits in the proxy, pretty much what you asked for:

- `max_concurrent_requests` (global + per-server), `queue_size`, `queue_timeout` in `mcp_config.json`
- extra requests wait in a bounded FIFO queue; when it's full (or waits too long) they're rejected with a clear "server busy, retry" tool result — REST gets 429 + Retry-After
- the limit sits at the single choke point all tool calls go through, so stdio upstreams are covered too (including calls made from `code_execution`)
- `0` = unlimited, so nothing changes unless you opt in; hot-reloadable so you can tune limits without restarting

That should remove the need for the nginx workaround, especially for stdio servers. Will post the PR here when it's up.

Dumbris:
Shipped in [v0.54.0](https://github.com/smart-mcp-proxy/mcpproxy-go/releases/tag/v0.54.0): native concurrency limits, exactly as discussed —

```json
{
  "max_concurrent_requests": 50,
  "queue_size": 100,
  "queue_timeout": "30s",
[...]

PR #959, opened 2026-08-07
feat(upstream): request queueing + per-upstream concurrency limits (spec 093)

Dumbris:
## Summary

Spec for native concurrency limiting decided in the #955 investigation (Option A of the decision report, included under `docs/research/`):

- **Per-server + global `max_concurrent_requests`**, bounded FIFO `queue_size`, `queue_timeout` — configured in `mcp_config.json`, per-server overrides, env overrides, hot-reloadable. **0 = unlimited: pure opt-in, zero behavior change on upgrade.**
- **Placement**: the single managed-client choke point, so every origin is covered — MCP variants, legacy call, direct routing, REST, and crucially the `code_execution` and activity-replay paths that bypass the manager layer (verified in the audit).
- **Shed semantics**: MCP `tools/call` → `isError:true` retryable tool result; REST → 429 + Retry-After; activity log status `rejected`; rejection counters + queue-depth metrics.
- Works identically in personal and server editions (all users share one managed client per upstream, so per-server limits bound aggregate multi-user load).

Related #955

## Review notes

Spec only — no code. Key verified findings: all four dispatch paths converge on one function while code_execution/replay bypass `Manager.CallTool`; stdio upstreams are genuinely multiplexed by mcp-go (limit ≠ 1); queue wait deliberately does not consume `call_tool_timeout`; `x/sync` is already a dependency.

Open maintainer decisions listed in the report (stdio default value, -32029 protocol-error convention, per-user fairness timing, health integration).


Dumbris:
## Implementation pushed — spec is now fully implemented

7 commits on top of the spec (limiter package → config surface → enforcement → shed semantics → REST exposure → e2e coverage).

**What landed**
- `internal/upstream/limiter`: two-tier admission limiter with shared cross-generation occupancy, FIFO, one absolute queue deadline, typed `LimitError{Scope, Reason, Server, RetryAfter}`, lifecycle tombstones (no admit-after-disable race), retire-after-drain.
- Config: three scopes per spec FR-020 (global aggregate / `server_concurrency_defaults` / per-server tri-state overrides), validation, hot-reload with atomic generations, env overrides for the global scope, swagger + docs.
- Enforcement at `managed.Client.CallTool` — covers all origins incl. `code_execution` and activity replay; **`Manager.CallTool` no longer holds `m.mu` across the upstream call** (FR-008); replay's execution timeout now starts after admission (FR-005).
- Shed semantics: `isError` tool result (scope-aware message), REST 429 + Retry-After with typed-error preservation through `CallToolDirect`, activity status `rejected` propagated through storage, API filters, summaries, usage aggregates, CLI, TUI, generated contracts + Web UI types, and Prometheus counters + queue-depth gauge.

**Local verification (all on this branch)**
- `go test -race ./internal/...` — 60 packages ok, 0 fail
- `go test -tags server ./internal/serveredition/... -race` — 7 ok
- `golangci-lint` (v2, CI config) — 0 issues
- `make swagger` — clean diff · frontend build + vitest: 588 tests pass · type-check clean
- New `internal/server/e2e_concurrency_limits_test.go` against a live slow upstream: max=1/queue=1 → call 2 queues then succeeds, call 3 sheds <100 ms, code_execution path limited too, upstream peak concurrency asserted =1; queue-timeout shed at ~250 ms. `-count=5 -race` no flakes.
- `./scripts/test-api-e2e.sh` — **65/65 pass**

Known scope calls: no Web-UI form control for per-server limits yet (REST + types are wired; no FR requires the form), CLI debug client documented out of scope.

Related #955

Dumbris:
## Cross-model review status: 5/5 rounds used — 2 findings outstanding

Codex reviewed this PR through 5 fix→re-review rounds (3 on the spec, 2 on the implementation; 18 code findings fixed and re-verified). The final round left two narrow findings, documented here per the review-cap policy:

1. **P1 — FIFO window during cap-raise reloads**: `Registry.publishLocked` stores the new generation before re-granting queued waiters; in that window a fresh admission can observe the raised cap's free capacity and jump ahead of existing waiters (`registry.go:270`, `limiter.go:245` fast path doesn't consult `l.waiters`). Fairness-only, bounded to the reload window; no deadlock/cap violation.
2. **P2 — queue-budget fallback not resolved per scope**: a capped scope with `queue_timeout: 0` is ignored when the other scope supplies a positive timeout (should be `min(30s fallback, other)`); the rejection's `RetryAfter` also reports 0 for that scope (`registry.go:307`, `limiter.go:344`).

Both are small, localized fixes in `internal/upstream/limiter`. Awaiting maintainer direction on whether to continue past the round cap.

Related #955

Dumbris:
✅ **Cross-model review complete — CLEAN.** The two outstanding findings were fixed in `30ce948fd` with maintainer authorization for a final round: FIFO is now preserved across cap raises (fast path yields to a non-empty wait queue; raise-created capacity drains to waiters first) and queue budgets resolve per scope (min of each active scope's effective timeout; Retry-After reports the shedding scope's value). Codex verified with 5,000-iteration FIFO stress + 200 regression passes: no new defects.

Related #955

PR #973, opened 2026-08-11
feat: Aggregate upstream MCP prompts through prompts/list

### LINKED ISSUE #972: [Feature]:  Aggregate/proxy upstream servers' MCP prompts through prompts/list
### Problem / motivation

  mcpproxy supports the MCP prompts capability, but only exposes two built-in, hardcoded prompts (setup-new-mcp-server, troubleshoot-mcp-server) via registerPrompts() in
  internal/server/mcp.go. Upstream MCP servers connected through mcpproxy that advertise their own prompts capability are not surfaced to the client at all — mcpproxy's upstream
  client code detects and logs whether an upstream advertises Capabilities.Prompts (internal/upstream/cli/client.go:208), but nothing forwards those prompts through mcpproxy's
  own prompts/list/prompts/get. This is inconsistent with how mcpproxy already handles tools/list, where tools from all connected upstream servers are aggregated and re-exposed.
  As a result, any workflow or client relying on prompts loses access to upstream-defined prompts as soon as it goes through the proxy.

### Proposed solution

  Extend mcpproxy's prompt handling to aggregate prompts from connected upstream servers, the same way tools are aggregated today:

  - On prompts/list, merge the two built-in prompts with the prompts advertised by every connected upstream server whose Capabilities.Prompts is non-nil, applying the same
  namespacing/prefixing convention used for tools to avoid name collisions across servers.
  - On prompts/get, resolve the requested (possibly prefixed) prompt name to its owning upstream server and forward the request, returning that server's GetPromptResult
  unchanged.
  - Respect the existing enable_prompts config flag as the master switch for this behavior.

  - Opt-in per-server aggregation flag: Add a per-upstream-server config option (e.g. expose_prompts: true) instead of aggregating by default for every server that advertises the
  capability. This gives finer control and avoids surprises if two servers register colliding prompt names, but adds config surface area and an extra step users must remember
  when adding a new server. Could be layered on top of the default-aggregation approach later if collisions or unwanted exposure become a real problem.

### Alternatives considered

do nothing

### Which edition?

Both / not sure

### Pre-submission checklist

- [x] I searched existing issues and discussions for a similar request

nlaurance:
## Summary
- Aggregates prompts from connected upstream servers that advertise `Capabilities.Prompts` into mcpproxy's own `prompts/list`/`prompts/get`, alongside the existing built-in prompts.
- Adds a per-server `expose_prompts` override so a server can be excluded from aggregation even if it advertises the capability.
- Scoped to the default (`retrieve_tools`-mode) server only, per the design spec's non-goals.
- Also fixes two gaps caught during verification: `ExposePrompts` wasn't persisted through the BBolt storage layer (would have been lost across a restart), and `RefreshPrompts`'s upstream fan-out had no timeout (could block the shared routing-refresh goroutine on a hung upstream).

Closes #972

## Test plan
- [x] `go build ./... && go vet ./... && go test ./internal/... -race` (4541 passed; one pre-existing, unrelated failure in `launcher.TestWaitForURL_InfersDefaultPort`, confirmed pre-existing on the base branch)
- [ ] Manual: connect a real upstream advertising prompts and confirm `prompts/list`/`prompts/get` surface it correctly


Dumbris (changes requested):
## Review: aggregate upstream MCP prompts through prompts/list

Thanks @nlaurance — this is a well-structured contribution. We checked it out locally (head 821428d), built both editions, ran the new tests under `-race`, ran the strict CI lint config, and QA'd the running binary against a live prompts-serving upstream (`@modelcontextprotocol/server-everything`) over both transports. The aggregation core is genuinely solid: correct `server__prompt` naming with no colon leak, argument passthrough works, quarantined/disabled servers are excluded with **no leak on either list or get**, graceful all-upstreams-down behavior, clean `enable_prompts:false` gating, and `notifications/prompts/list_changed` fires on refresh. Test discipline is unusually good for an external PR.

That said, live QA found one blocking issue the unit tests can't see, plus a few smaller ones.

### Blocking

1. **P1 — Prompts are unreachable over the Streamable HTTP `/mcp` endpoint in every routing mode.** `RefreshPrompts`/`AddPrompt` register only on `p.server` (`internal/server/mcp_routing.go:720`, `mcp.go:1080`), but `/mcp` serves `GetMCPServerForMode(cfg.RoutingMode)` (`internal/server/server.go:774`), which returns `callToolServer`/`directServer`/`codeExecServer` — none created with `WithPromptCapabilities` (`mcp_routing.go:563-602`). Since `config.Validate()` normalizes `routing_mode` to `retrieve_tools` (`internal/config/config.go:2430`), the `p.server` fallback is never hit over HTTP. Verified live: `initialize` on `/mcp` advertises no `prompts` capability and `prompts/list` returns `-32601 "prompts not supported"`, while the same binary over **stdio** serves all aggregated prompts correctly. As it stands the headline feature only works for stdio deployments. Fix: register prompts (and the prompts capability) on the routing-mode server instances actually served at `/mcp`.

2. **P2 — `expose_prompts` cannot be toggled at runtime by any path.**
   - REST: `PATCH /api/v1/servers/{id}` with `{"expose_prompts": false}` is rejected with `{"success":false,"error":"No fields to update"}` — `MergeServerConfig` (`internal/config/merge.go:159+`) has no `ExposePrompts` branch (only `CopyServerConfig` was updated). Please add the non-nil-pointer merge branch like `InitTimeout`/`MaxConcurrentRequests`, plus a `merge_test.go` case.
   - File hot-reload: flipping `expose_prompts` alone bumps the config version but emits no `servers.changed`, `RefreshPrompts` never runs, and the core client's cached ServerConfig stays stale — the toggle only takes effect when coupled with a reconnect-forcing change (enabled/quarantine flip) or restart. It needs to be wired into config change detection too.

3. **P2 — Missing SSE request serialization in `core.Client.ListPrompts`/`GetPrompt`.** `ListTools` (`internal/upstream/core/client.go:300-305`) and `CallTool` (`:400-401`) take `c.sseRequestMu` for SSE transports because concurrent requests cause response-delivery failures; the new `core/prompts.go` methods skip this, so a prompts refresh racing a tool call against an SSE upstream reintroduces exactly that failure mode. Please mirror the `ListTools` pattern.

4. **P3 — gofmt:** `gofmt -l` flags all four new files (`internal/upstream/core/prompts.go`, `core/prompts_test.go`, `managed/prompts.go`, `managed/prompts_test.go` — missing trailing newline). Mechanical.

### Recommended (non-blocking)

5. `Manager.GetPrompt` (`internal/upstream/manager_prompts.go`) lacks the `Enabled`/`Quarantined` guards `Manager.CallTool` enforces (`manager.go:1220`). QA showed deregistration currently masks this (`prompts/get` on a quarantined server returns "prompt not found"), but there's a race window between quarantining and the `servers.changed`-driven refresh during which the proxy would still forward. Defense-in-depth: mirror CallTool's checks.
6. Pagination: `core/prompts.go` fetches only the first page (`NextCursor` ignored). Consistent with `ListTools` today, so fine as a follow-up — a `// TODO` would help.
7. Upstream `notifications/prompts/list_changed` is not consumed (re-aggregation only on `servers.changed`) — worth a line in the design doc's non-goals.
8. Docs: `docs/configuration.md` has no `expose_prompts` entry. Test nit: `disableOAuthForTest` in `core/prompts_test.go` hand-rolls env save/restore; `t.Setenv` does it in one line.

### Verified locally

`go build ./...` + `-tags server` both pass · prompt unit tests green under `-race` · `golangci-lint` v2 CI config: 0 issues on touched packages · no merge conflicts with main · live QA evidence: stdio prompts/list returns `everything__args-prompt`, `everything__completable-prompt`, `everything__resource-prompt`, `everything__simple-prompt` + 2 built-ins; `prompts/get everything__args-prompt {city, state}` echoes arguments; after quarantine: list drops to built-ins and get returns `-32602` (no leak).

Happy to merge once items 1-4 are addressed — the core design won't need to change.

Dumbris (approved):

Dumbris:
Pushed a follow-up commit (`326c5bdb8`) that makes prompt aggregation **safe by default** and closes the two highest-severity gaps from review. CI is green.

**What changed**

- **Aggregation is now opt-in.** New `aggregate_upstream_prompts` config flag, default `false`. It gates *only* the upstream aggregation in `RefreshPrompts`; `enable_prompts` still governs the built-in prompts + the prompts capability, so the safe built-ins keep working out of the box and users opt into upstream aggregation deliberately. Wired through config default, CLI flag, viper, hot-reload change detection, telemetry, OpenAPI, and docs.
- **The switch is a live read.** `RefreshPrompts` reads the hot-reloaded config snapshot (not the boot-time copy) and re-runs on `config.reloaded`, so toggling the flag takes effect without a restart.
- **Scope enforcement on aggregated prompts.** Registered `mcpserver.WithPromptFilter` on every routing-mode server. mcp-go applies it to both `prompts/list` and `prompts/get`, so an agent token or profile session scoped to one server can no longer list or fetch another server's prompts. Built-in prompts stay visible to everyone.
- **`expose_prompts` enforced on get.** `core.Client.GetPrompt` now checks `expose_prompts` (and the prompts capability), matching `ListPrompts` — an opted-out server's prompt content can no longer be fetched by name.

Tests added: default-off + live-read regressions, aggregation-disabled (built-ins only), the `WithPromptFilter` scope matrix, `core.GetPrompt` opt-out, and hot-reload change detection.

**Deferred to a follow-up PR** (tracked; feature stays off by default until then): scanning/sanitisation of upstream prompt content (parity with the tool TPA + output-sanitisation path), activity logging for `prompts/get`, a size/count cap + pagination, consuming upstream `prompts/list_changed`, per-server REST/UI wiring for `expose_prompts`, and name-collision/`:`-in-name hardening.

PR #1015, opened 2026-08-21
fix(tray): forward --listen to the spawned core process

electrolobzik:
## Bug

`mcpproxy-tray serve --listen 0.0.0.0:8181` silently drops the `--listen` flag on macOS (reproduced on tray v0.43.0):

- The tray log shows the core spawned as `/bin/zsh -l -c exec '/mcpproxy' 'serve'` — no `--listen`.
- `lsof` confirms the core then listens on the config-file value (default `127.0.0.1:8080`); nothing ever opens the advertised port, so any tooling that probes it concludes the daemon is down.

## Root cause

`cmd/mcpproxy-tray/main.go` never parses its own command line — `serve --listen …` is ignored entirely. On macOS `resolveCoreURL()` prefers the unix socket for tray↔core communication (#102), and `buildCoreArgs()` only derives `--listen` for TCP/HTTP core URLs, so on the socket path the core is spawned with no `--listen` at all and falls back to the config default.

## Fix

- New `trayListenFromArgs()` parses `--listen`/`-l` (both space and `=` forms) from the tray's argv; other args remain ignored as before. Malformed values (dangling flag, empty `=` form, value starting with `-`) are skipped and scanning continues to the first valid value.
- `buildCoreArgs()` now always forwards an explicit CLI `--listen` to the spawned core's argv — including when the tray talks to the core over the socket/pipe, since the socket only covers tray↔core communication while the core must still open the advertised TCP address. Both the login-shell-wrapped spawn and the direct-exec fallback consume the same argv.
- URL/env-derived `--listen` behavior for TCP endpoints is unchanged.

## Tests

- `TestTrayListenFromArgs` covers both flag forms plus the malformed-input regressions (`--listen --config path`, `--listen= --listen :8181`, dangling `-l`, continue-scan to a later valid `--listen=:9090`).
- `TestBuildCoreArgs_ForwardsCLIListenOverSocketEndpoint` — regression: CLI listen forwarded over a socket endpoint; no `--listen` on socket endpoints without the flag.
- CI on our fork built all platform artifacts (incl. darwin-arm64) green: https://github.com/HaloCollar/mcpproxy-go/pull/7

### DIFF
--- a/cmd/mcpproxy-tray/main.go
+++ b/cmd/mcpproxy-tray/main.go
@@ -47,6 +47,7 @@ var (
 	defaultCoreURL   = "http://127.0.0.1:8080"
 	errNoBundledCore = errors.New("no bundled core binary found")
 	trayAPIKey       = ""                  // API key generated for core communication
+	trayCLIListen    = ""                  // --listen from the tray's own command line (set in main)
 	shutdownComplete = make(chan struct{}) // Signal when shutdown is complete
 	shutdownOnce     sync.Once
 )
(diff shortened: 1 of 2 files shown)


Dumbris:
Merged — thanks Roman! Great catch that the tray never parsed its own argv, so `mcpproxy-tray serve --listen 0.0.0.0:8181` was silently dropped on the macOS socket path and the core quietly fell back to 127.0.0.1:8080; the fix is tightly scoped and the malformed-flag test cases are a nice touch. 🙏

Also useful method is to set using env var https://docs.mcpproxy.app/configuration/environment-variables#setting-tray-variables-on-macos

PR #1023, opened 2026-08-22
fix(tray): slim the menu + represent code_execution by its sub-calls

Dumbris:
Owner feedback on the tray menu (scrolling appeared) and on code_execution's place in the glance.

## Menu length
- Removed the grey "N quarantined server(s)" info line (Needs Attention already carries it)
- Removed "Report an Issue…" (ProjectLinks.issues kept — mirrored in branding + Web UI)
- "Add Server…" moved into the Servers submenu as its first item (⌘N still works; falls back to top level only when there are zero servers, so the empty-state user can still reach it)
- New MenuStructureTests pin the contract through the real rebuildMenu()

## code_execution is an internal primitive
The glance now shows the real upstream calls a script made — not the wrapper row: `code_execution` removed from the glance discovery set (Swift GlanceSelection + AppState live-count mirror + Go usage aggregate, in lockstep). A FAILED code_execution still rows (it has no children to speak for it). `usageAdmissionVersion` bumped 3→4 so existing snapshots rebuild under the new rule. Parity test rewritten as an independent two-stage oracle; spec 069 usage contract amended. The Web UI activity list intentionally still shows parent rows (drill-down home).

Tests: swift build + 467 filtered Swift tests green (incl. 5 new MenuStructureTests), go runtime suite + lint 0 issues.


Dumbris:
Cross-model review (opencode/gpt-5.6-sol) returned 2 P2s, both pre-existing trade-offs rather than regressions of this PR: (1) rejected concurrency sheds are glance rows but stay out of the timeline — the deliberate spec-093 contract, open product question since #1016; (2) failed call_tool_* wrappers reaching the glance via SSE — defended in practice by the server-side default filter on poll and the rule-4 request-id collapse, so no double rows render. Documented here rather than changed.

PR #1052, opened 2026-08-25
fix(webui): make the Activity and Usage numbers say what they mean

Dumbris:
Fixes the Activity & Usage cluster of the Web-UI UX audit (#1046): **F2, F5, F22, F23, F25, F26, F27**, plus the Activity/Usage item in F36. Builds on the `storage.CountsAsCall` contract landed by #1048.

Every finding below was verified live with Playwright against an isolated instance (scratch `--config` + `--data-dir`, port 18491, four servers, ~120 activity rows of generated traffic).

---

### F2 — the status tiles did not sum

`Total (24h) 42 · Success 15 · Errors 4 · Blocked 0 · Rejected 0` — four buckets adding to 19 under a denominator of 42.

The four are the tool-call status vocabulary, but the activity log is wider than tool calls: a quarantine change stores its **action** in `status` (`tool_auto_approved`), a policy decision its **verdict** (`allow`). Those rows were in the total and in no bucket.

`ActivitySummaryResponse` now carries `other_count`, computed as the `default` branch of the same switch, so the partition holds **by construction**. The tile list moved into one pure function (`statusBucketTiles`), an "Other / internal" tile renders the residual with a tooltip explaining what it holds, and it filters like every other tile. A Go test and a Vitest test both assert the sum; live: `23 + 6 + 2 + 0 + 23 = 54`.

### F5 — repeats did not group, and every row said "Success"

Consecutive rows that agree on everything the collapsed line prints now fold into one expandable run. The rule that makes the compression safe is that the run identity covers **every field the lead row displays on the members' behalf** — so status is in it (a run can never mix outcomes; `echo ×12` can never hide the one call that failed), and so are the intent operation type and the sensitive-data severity and count. The intent *reason* is the one displayed field deliberately left out — agents reword it call to call, which would defeat the compression on exactly the logs that need it — so a run whose members disagree shows a `+…` marker instead.

Folding is on by default, remembered per browser, and disabled whenever the table is sorted by something other than time — adjacency only means "repeated" in time order. The paginator counts runs and names the rows behind them.

Successes no longer print "Success" at all: the word is `sr-only`, so the marked rows in the Status column are exactly the ones needing a human. Live: 12 identical echo calls with a failure at position 8 rendered as `echo ×9` / `doesnotexist` (the only red chip) / `echo ×7`.

### F22 — the usage charts invented tools and quantized latency

**Invented tools.** `everything:doesnotexist` and `broken-remote:whatever` were charted as first-class tools with 100% error rates. Those names came from failed calls — an agent's typo, a server that was never reachable. A name that has never once completed a call is excluded from *Calls per tool* and *Token sinks* (an error message is a response with bytes, which is how a name that only ever failed ranked as a "token sink"), named in a footnote, and still charted under *Errors & latency*, where it belongs. The "active tools" tile counts the same population, and admits the folded tail when the server truncates the list. Self-correcting: one success and it reappears.

**Latency.** Every row of the p50/p95 table read exactly `10 ms` while the Activity Log showed 3/4/5 ms for the same calls. A histogram percentile is the *upper bound of a bucket*, and the first bound was 10 ms — the band local MCP servers live entirely inside. `latencyBucketBoundsMs` gains `1, 2, 5`, and `Percentile` now also reports whether it landed in the unbounded overflow bucket, so the table renders `≤ 5 ms` or `> 10 s` and can never be read as a stopwatch again. `usageAdmissionVersion` is bumped to 5 (the stamp now covers bucket layout, not just admission) so persisted snapshots rebuild rather than carry indices that changed meaning.

**Axes.** Every chart names what it measures: *Calls (cumulative)*, *Total response bytes*, *Error rate (% of calls)*, plus a line separating the error bars from the latency table beneath them.

Live: `≤ 10 ms / ≤ 5 ms / ≤ 5 ms / ≤ 1 ms / —` where the whole column used to read `10 ms`.

### F23 — the token-savings headline did not move

**It is not supposed to.** Tracing the aggregation path: `CalculateTokenSavings` is a *structural* estimate over the current tool catalog — what it would cost to put every upstream tool in the agent's context versus what `retrieve_tools` returns for one query. It moves when servers or tools change, never per call. The number was right; the label ("Tokens saved", beside per-window call counts) said cumulative-savings-so-far.

Relabelled to *Tokens saved per request* / *smaller tool context per request*, given the derivation on hover, and the Dashboard chip is now a control that opens the Token Savings breakdown that computes it.

### F25 — the drawer for internal events was a dead end

A `security_scan` row now renders its verdict, its findings by severity, and two ways out (the server, and the Security page). A session id is now a control rather than plain text: *Filter this log* (preferring the Spec 082 work session) and *Open Sessions* — Sessions has linked into Activity since 082 and nothing linked back.

Building it uncovered why the drawer had nothing to show. **The findings summary never reached the record at all**: `publishScanSettled` puts a `map[string]int` in the payload and `handleSecurityScanSettled` read it with a `map[string]interface{}` assertion that never matched, so every `security_scan` record ever written has `metadata: null`. Fixed at the read, keeping an **empty** rollup distinct from an **absent** one — "the scanners found nothing" and "this record has no summary" are opposite claims, and the drawer must not print the first when it only knows the second. Verified live: a freshly triggered scan now writes `{"findings_summary": {}}`.

> Deviation from the suggested fix: the drawer does **not** link to `/security/scans/:jobId`. The settled-scan event is debounced across scanners and carries no job id, so there is no single report to link to without threading one through the debouncer — out of scope for a UX pass. It links to the Security page and the server instead.

### F26 — Intent and Sensitive were unlabelled emoji

The Intent column paints the operation word beside the glyph (`📖 read`, not `📖`), the Sensitive badge gains an accessible name, and the filter panel carries a visible legend for both vocabularies. Colour + glyph is no longer the sole encoding (WCAG 1.4.1).

### F27 — an ambiguous denominator and a double count

*Connected* read `50% online` for 2 of 4 servers, one switched off and one quarantined; a server that is not enabled cannot be online. The tile now states its denominator: **"2 of 3 enabled online"**. (Quarantined servers stay in it: `lifecycle.go` keeps them **connected** for security inspection, so they are supposed to be up.)

`serverCount.disabled` is a new count asked directly — switched off **and** not quarantined — replacing `total − connected − quarantined`, which was wrong twice: it double-counted a server that was both (the Dashboard rail read one server as two problems) and it labelled an *enabled but unreachable* server "disabled", hiding the actual fault behind an administrative word.

### F36 — `View Activity` clipped

`whitespace-nowrap`. Verified: `scrollW == clientW`, one line.

---

### Tests

- **Go**: the summary partition; four `Percentile` cases (sub-10 ms resolution, the overflow floor, an empty histogram, the existing tail-latency case); the p95 overflow tie-break; and the scan-payload reader, including the empty-vs-absent distinction.
- **Vitest**: 4 new spec files, 39 new tests — the status partition, run grouping (including the "never mixes outcomes" property and the seven things that must never be folded away), scan-rollup ordering and presence, unresolved-tool partitioning, latency-bound formatting, and a component pass over the table (folding, expansion, the blank success cell, the intent word, the legend, the Other tile, the scan drawer).
- Full suite: 742 frontend tests, `vue-tsc` clean, `golangci-lint` v2 clean, both editions build, `internal/{httpapi,runtime,contracts,storage}` green under `-race`.
- **Not run:** `scripts/test-api-e2e.sh` — its cleanup blanket-pkills cores it does not own, and other sessions had cores live.

### Cross-model review

Four rounds (opencode / gpt Sol). Nine findings fixed — see the [review note](#issuecomment-5410340877) for the full list, including the two that were not in the audit (the dropped scan summary above, and three ways a run could fold away a field it was displaying). Three findings were checked and rejected, with reasons.

### Two accepted trade-offs, flagged for a human

1. **"Never completed a call" also excludes a *real* tool failing 100% in the window** (in the live run, `everything:add`). Intended: it is not evidence of use, it is named in the footnote, and it stays in the errors chart. This is the judgement call most worth a second opinion.
2. **The "Other / internal" tile counts the whole 24 h server-side while its filter narrows the 200 loaded rows.** True — and true in exactly the same way for Success / Errors / Blocked / Rejected, before this PR and after it. Making the tile row server-filtered is a real improvement and a different change from making it add up.

### API surface

Additive: `ActivitySummaryResponse.other_count`, `UsageToolStat.p50_exceeds` / `p95_exceeds`. `oas/` regenerated.


Dumbris:
## Cross-model review (opencode / gpt Sol) — 4 rounds

Round 1 (backend + shared utils) and rounds 2–3 (views, then the whole diff) found nine issues; all nine are fixed in commits 2–4. The two that mattered most were not in the audit at all:

**The scan findings summary never reached the record.** `Runtime.publishScanSettled` puts a `map[string]int` in the event payload; `handleSecurityScanSettled` read it with a `map[string]interface{}` type assertion, which never matched. Every `security_scan` activity record ever written has `metadata: null`. Nothing noticed because nothing rendered it — until F25's drawer did. Fixed at the read, with the **empty vs absent** distinction preserved: a scan that found nothing reports `{}`, and the drawer may say "no findings"; a record with no rollup gets told it is not the same as a clean result. Verified live — a freshly triggered scan now writes `{"findings_summary": {}}`.

**Three ways a run could fold away what the collapsed row was displaying.** The intent operation type was not in the key (a run of `call_tool_write` calls could fold under a "read" lead — the two produce records identical in type, server, tool and status); only the sensitive-data *boolean* was, though the badge prints the severity glyph and the count (a critical detection could fold under a low-severity lead); and the key was a space-joined string over free-form fields. Four regression tests, one per hazard.

Also fixed: `currentPage` was never clamped, so folding repeats on while on a later page rendered an empty table; expanded runs and the page now reset on sort/session/grouping changes too; exporting from the "Other / internal" view sent a synthetic status to an exact-match endpoint and came back empty; sorting by p95 could not tell a tool *bounded* at the last histogram bucket from one that ran *past* it, which matters under top-N truncation; and a failed scan's retained findings are now shown labelled as partial rather than hidden.

### Rejected, with reasons

- **"The Servers enabled denominator should exclude quarantined servers."** Raised twice; wrong. `internal/runtime/lifecycle.go` keeps a quarantined server **connected** for security inspection, so it is supposed to be up and belongs in the denominator.
- **"All-error rows are treated as unresolved."** That is the design, not a defect — see the trade-off note above. A real tool failing 100% in the window is excluded from *use* charts, named in the footnote and still charted under errors.
- **"'Other / internal' filters the 200 loaded rows while its tile counts the whole 24h window."** True, and true in exactly the same way for the Success / Errors / Blocked / Rejected tiles, before this PR and after it — every status tile is a server-side count driving a client-side filter. Making the tile row server-filtered is a real improvement and a different change from making it add up.

Round 4 returned only the last two, i.e. no new defects.

PR #1057, opened 2026-08-25
fix(security): redact custom credential header names

neylwalecki:
# Pull Request

## Description

Known credential headers such as `Authorization` and `X-API-Key` are redacted in server listings and logs, but custom names such as `access_token`, `X-Custom-Secret`, and `X-Client-Credential` can currently expose an opaque value unchanged. The fallback value scanner cannot identify a secret when the value itself has no recognizable `token=` or `secret=` prefix.

This change centralizes header-name classification in `isSensitiveHeaderKey`, preserves the existing exact-name list, and additionally recognizes delimiter-separated credential segments. The same classifier is used by both `RedactHeaders` and `RedactStringHeaders`.

Negative regression coverage keeps non-sensitive names such as `X-Author-ID` and `X-Monkey-ID` readable, avoiding the false positives introduced by raw substring matching.

Related context: #872 and #875.

## Testing

- [x] All existing tests pass

Behavioral verification:

- The new regression tests fail before the fix because custom credential values remain in plaintext.
- Positive cases cover `access_token`, `X-Custom-Secret`, and `X-Client-Credential`.
- Negative cases cover `X-Author-ID` and `X-Monkey-ID`.
- API E2E completed with 65 passed, 0 failed, and 0 skipped.

### DIFF
--- a/internal/oauth/logging.go
+++ b/internal/oauth/logging.go
@@ -580,7 +580,7 @@ func UnmaskHeaders(incoming, stored map[string]string) map[string]string {
 // renders one (key, value) pair; reused by UnmaskHeaders to recognise echoed
 // masks.
 func maskedHeaderValue(key, value string) string {
-	if sensitiveHeaders[strings.ToLower(key)] {
+	if isSensitiveHeaderKey(key) {
 		return MaskValue(value)
 	}
 	return RedactSensitiveData(value)
--- a/internal/oauth/logging.go
+++ b/internal/oauth/logging.go
@@ -25,6 +25,42 @@ var sensitiveHeaders = map[string]bool{
 	"proxy-authorization": true,
 }
 
+// sensitiveHeaderSegments catch credential-bearing custom header names that
+// cannot be enumerated exhaustively (for example access_token or
+// X-Client-Credential). Matching complete delimiter-separated segments avoids
+// false positives such as X-Author-ID and X-Monkey-ID.
+var sensitiveHeaderSegments = map[string]bool{
+	"auth":          true,
+	"authorization": true,
+	"bearer":        true,
+	"cookie":        true,
+	"token":         true,
+	"secret":        true,
+	"key":           true,
+	"apikey":        true,
+	"password":      true,
+	"passwd":        true,
+	"credential":    true,
+	"private":       true,
+	"session":       true,
+}
+
+var headerNameSegmentPattern = regexp.MustCompile(`[a-z0-9]+`)
+
+func isSensitiveHeaderKey(name string) bool {
+	if sensitiveHeaders[strings.ToLower(name)] {
+		return true
+	}
+
+	for _, segment := range headerNameSegmentPattern.FindAllString(strings.ToLower(name), -1) {
+		if sensitiveHeaderSegments[segment] {
+			return true
+		}
+	}
+
+	return false
+}
+
 // Sensitive parameter names in request bodies or URLs.
 var sensitiveParams = []string{
 	"access_token",
@@ -92,8 +128,7 @@ func RedactHeaders(headers http.Header) map[string]string {
 	redacted := make(map[string]string)
 
 	for key, values := range headers {
-		lowerKey := strings.ToLower(key)
-		if sensitiveHeaders[lowerKey] {
+		if isSensitiveHeaderKey(key) {
 			redacted[key] = "***REDACTED***"
 		} else {
 			// Join multiple values and redact any sensitive data within
@@ -129,8 +164,7 @@ func RedactStringHeaders(headers map[string]string) map[string]string {
 	}
 	redacted := make(map[string]string, len(headers))
 	for key, value := range headers {
-		lowerKey := strings.ToLower(key)
-		if sensitiveHeaders[lowerKey] {
+		if isSensitiveHeaderKey(key) {
 			redacted[key] = MaskValue(value)
 		} else {
 			redacted[key] = RedactSensitiveData(value)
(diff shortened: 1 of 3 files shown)


Dumbris:
Thanks for this — the segment-matching idea is exactly right, and the `X-Author-ID` / `X-Monkey-ID` cases in the tests show you thought about the substring false positives that a naive `strings.Contains` would hit. Build, `go test -race ./internal/oauth/...` and golangci-lint v2 are all green on the branch.

There is one blocker before this can land, though: widening `RedactStringHeaders` without widening its inverse breaks the mask round-trip and can persist the mask **over the real secret**.

`maskedHeaderValue` is documented as "the single source of truth for how `RedactStringHeaders` renders one (key, value) pair" and is what `UnmaskHeaders` uses to recognise an echoed-back mask — but it still consults the narrow `sensitiveHeaders` map:

```go
func maskedHeaderValue(key, value string) string {
	if sensitiveHeaders[strings.ToLower(key)] { // <- not isSensitiveHeaderKey
		return MaskValue(value)
	}
	return RedactSensitiveData(value)
}
```

So for a header that is newly caught by the segment matcher, the read path emits `••••xx (N chars)` but the write path no longer recognises that string as the mask, and stores it verbatim. Repro (fails on this branch, passes on `main`):

```go
func TestMaskRoundTrip(t *testing.T) {
	stored := map[string]string{"X-Custom-Secret": "supersecretvalue"}
	read := RedactStringHeaders(stored)                       // "••••ue (16 chars)"
	incoming := map[string]string{"X-Custom-Secret": read["X-Custom-Secret"]}
	out := oauth.UnmaskHeaders(incoming, stored)
	// got "••••ue (16 chars)", want "supersecretvalue"
	assert.Equal(t, stored["X-Custom-Secret"], out["X-Custom-Secret"])
}
```

That path is live in two places — `PATCH /api/v1/servers/{id}` (`internal/httpapi/server.go:2043`) and the `upstream_servers` MCP patch (`internal/server/mcp.go:5146`) — and an MCP client that reads a server and echoes the whole `headers` map back is the normal shape of an update there. The result is a destroyed credential rather than a leaked one, which is the failure mode the earlier rounds on this file (see the `UnmaskEnvValues` doc comment) were specifically hardening against.

The fix should be a one-liner — switch `maskedHeaderValue` to `isSensitiveHeaderKey(key)` — plus a regression test asserting `UnmaskHeaders(RedactStringHeaders(stored), stored) == stored` for a segment-matched custom header, so the read/write pair stays pinned together. The `env` side already gets this right (`maskedEnvValue` calls `isSensitiveEnvKey`), which is a good model to mirror.

Two non-blocking notes while you're in here:

- The segment pattern `[a-z0-9]+` glues trailing digits and plurals onto the word, so `X-Tokens`, `X-Keys` and `X-Secret2` still slip through. Not a regression (they leak today too), but if you want to close it, matching `[a-z]+` and letting digits act as delimiters — or checking a trimmed plural — would cover it.
- `session` and `private` will now also mask things like `X-Session-Id` and `Access-Control-Request-Private-Network`. That is a safe direction to err in, just worth being deliberate about since these values become unreadable in the Web UI server detail.

Happy to re-review as soon as the unmask side is updated.

Dumbris:
Pushed a maintainer commit (7a7407c) landing the round-trip fix from the review above — everything else is untouched, and the redaction scope stays exactly as you designed it.

The one-line change: `maskedHeaderValue` now calls `isSensitiveHeaderKey(key)` instead of consulting the narrow 8-entry `sensitiveHeaders` map, mirroring how `maskedEnvValue` already defers to `isSensitiveEnvKey`. That function is the write-path inverse `UnmaskHeaders` uses to recognise an echoed-back mask, and `unmaskMapValues` requires `rendered(key, storedValue)` to reproduce exactly what the read path emitted. Once `RedactStringHeaders` started masking custom credential headers, the read path emitted `••••ue (16 chars)` for e.g. `X-Custom-Secret` but the write path no longer recognised its own mask — so an unedited echo looked like a genuine edit and persisted the mask over the real credential, on both `PATCH /api/v1/servers/{id}` and the `upstream_servers` patch.

Also added a round-trip regression test (`TestUnmaskHeaders_RestoresEchoedMask_CustomCredentialHeaders`) covering `X-Custom-Secret`, `X-Client-Credential` and `access_token`; it fails with the one-line change reverted. Your existing tests all still pass.

neylwalecki:
Verified `7a7407c` locally. Aligning `maskedHeaderValue` with `isSensitiveHeaderKey` restores the read/write invariant, and `TestUnmaskHeaders_RestoresEchoedMask_CustomCredentialHeaders` now pins the round-trip for the widened header classification.

Local verification passed:

- `go test ./internal/oauth -run 'TestRedactHeaders|TestRedactStringHeaders|TestUnmaskHeaders_RestoresEchoedMask' -count=1 -v`
- `go test -race ./internal/oauth -count=1`
- Focused HTTP API and runtime redaction tests
- `git diff --check`

The PR Build rerun and the remaining CI checks are green as well. I am keeping the plural/digit variants and the broader `session` / `private` masking trade-off out of this PR so the current security fix stays narrow. Marking it ready for review.

Dumbris:
Merged — thank you for this: custom credential headers like `X-Custom-Secret` and `X-Client-Credential` no longer leak their values into server listings or logs, closing a real gap that our hardcoded eight-name list had been quietly missing. The detail I especially appreciated is that matching complete delimiter-separated segments instead of raw substrings sidesteps the classic false-positive trap — `X-Author-ID` and `X-Monkey-ID` stay readable — and you pinned that behaviour down with tests, which made the change easy to trust.

neylwalecki:
Really glad to see this merged — the round-trip review made the fix stronger. I appreciated the thoughtful collaboration and look forward to contributing again!

PR #1075, opened 2026-08-28
fix(tray): stop the permanent red badge, and shrink it to a corner dot

Dumbris:
# Pull Request

## Description

Reported symptom: after updating to v0.61.0 the tray showed a red dot that never cleared, and it looked too big.

Both halves turned out to be real, plus two more defects found while tracing it. The badge itself is new in v0.61.0 (#1055) — this state existed before, it was simply invisible.

### 1. Quarantined servers no longer tint the badge red

`worstDiagnosticSeverity` / `diagnosticCount` excluded OAuth-login-required servers but not quarantined ones. mcpproxy still attempts a connection to a quarantined server so the scanner can export its tool definitions, and the failed attempt leaves an error-severity diagnostic behind — while the *same payload* reports `health.level: "healthy"` / `admin_state: "quarantined"`.

That is why the reporter's screenshot showed a calm **yellow** dot in the menu header (drawn from `serversNeedingAttention`) under a **red** menu-bar dot (drawn from `worstDiagnosticSeverity`), off one payload. The user could not clear it without approving or disabling the server.

Both properties now share one `ServerStatus.isBadgeExempt` predicate so they cannot drift apart, and any future intentional non-connected state has exactly one place to be added.

### 2. The badge is a corner dot, not a glyph beside the icon

It was a full-size `"●"` plus a leading space in the button's `attributedTitle`, which widened the status item by roughly a character and read as a second icon rather than a badge on the first. It is now a 6pt dot in the icon's bottom-right corner.

It has to be a **subview**, not pixels composited into `button.image`: that image must stay `isTemplate` to track the light/dark menu bar and to invert while the menu is open, and template rendering forces every pixel monochrome — so a red dot drawn into the image comes back black. The view is hit-test transparent (otherwise it would swallow the clicks that open the menu wherever it overlaps the icon) and ignored by accessibility, since the button already publishes the whole state as its label.

`⏹` (core stopped) and `⚠` (core error) keep their full-size glyphs. Those mean "nothing is working" — worth the width, and the distinction would be lost if collapsed into the same dot.

### 3. Two of mcpproxy's own error messages no longer say "file a bug"

`command "npx" has no args` (emitted by our own pre-spawn validation in `connection_stdio.go`) and the legacy-SSE `4xx for initialize POST` both fell through to `MCPX_UNKNOWN_UNCLASSIFIED`, whose catalog entry offers a **"Report a bug"** button — for a config typo the user can fix in one line.

They now classify as `MCPX_CONFIG_INVALID_COMMAND` and `MCPX_HTTP_LEGACY_SSE`, with catalog entries and docs pages. The legacy-SSE rule is matched *before* the generic HTTP status-text fallback, which would otherwise claim it as a bare 403/404 and send the user hunting for a credential problem that does not exist.

### 4. One `ListTools` failure was three ERROR lines

`core.Client`, `managed.Client` and `Manager.discoverTools` each logged the same failed list operation at ERROR. On an install with several stdio servers the periodic sweep hit this roughly once a minute per server, indefinitely — measured at ~110 ERROR lines/min during startup and ~14/min at steady state, rotating `main.log` every couple of hours.

The two inner layers drop to Debug. The manager keeps its line at **Warn**: it deliberately skips `markSwept` so the next sweep retries, and a failure the code already plans to retry is not an error.

## Testing

- [x] All existing tests pass

Verified against a live isolated instance (`MCPPROXY_HOME=/tmp/mcpproxy-dot`, port 18477) seeded with both fixtures — a misconfigured `npx` server and an unreachable quarantined server:

- Tooltip read **"1 server error"** with *two* error-severity diagnostics present — the quarantined one correctly excluded.
- Status item `title: ""`, width unchanged at 36pt — the badge no longer widens the menu bar.
- The menu still opens through the dot overlay (hit-test passthrough works).
- The broken server reported *"This server's command has nothing to run — a package runner like npx or uvx needs the package name in `args`"* instead of "file a bug report".

Automated:

- `go build ./...` clean; `golangci-lint run --config .github/.golangci.yml ./...` → **0 issues**.
- Go tests pass across `internal/diagnostics`, `internal/upstream/...`, `internal/telemetry`.
- Swift: 1127 tests, 1 failure — `AppLifecycleTests.testTheSharedJournalNeverWritesToTheRealInstanceRootUnderTests`, which asserts `~/.mcpproxy/tray-lifecycle.jsonl` does not exist and fails on any machine with a running tray. Pre-existing and environmental, unrelated to this change.

New tests:

- Swift badge exemption, built from the verbatim `/api/v1/servers` payload shapes — including that a genuinely broken server *still* tints red, and that the spoken count agrees with the badge.
- Swift corner-dot geometry and colour, asserted by rendering the view offscreen and reading pixels back. The status item cannot be screenshotted without a Screen Recording grant, and "it looked right once" is not a regression test.
- Go: both new codes classify from the verbatim v0.61.0 error strings; ordinary HTTP statuses are unaffected; both codes are registered in the catalog.
- Go: one `ListTools` failure is reported by exactly one layer, not at ERROR, and is still reported at all.

## Notes for the reviewer

- **The corner dot has not been seen by a human.** Screen Recording is not granted in the session that built this, so it was verified by offscreen pixel rendering and by the status item's reported title/width rather than by eye. Worth one look after rebuilding the tray.
- `./scripts/test-api-e2e.sh` was **not** run: it blanket-`pkill`s mcpproxy cores and would have killed the reporter's live core mid-session. Worth running in CI or on a quiet machine.
- New docs pages are indexed in both `docs/errors/README.md` and `website/sidebars.js` — the file allowlist is a glob, but the sidebar is manual.

### DIFF
--- a/docs/errors/MCPX_CONFIG_INVALID_COMMAND.md
+++ b/docs/errors/MCPX_CONFIG_INVALID_COMMAND.md
@@ -1,0 +1,82 @@
+---
+id: MCPX_CONFIG_INVALID_COMMAND
+title: MCPX_CONFIG_INVALID_COMMAND
+sidebar_label: INVALID_COMMAND
+description: A stdio server's command and args cannot spawn anything — typically a package runner with no package to run.
+---
+
+# `MCPX_CONFIG_INVALID_COMMAND`
+
+**Severity:** error
+**Domain:** Config
+
+## What happened
+
+The server's `command` / `args` pair cannot start an MCP server. The common case
+is a package runner named with nothing to run:
+
+```jsonc
+{
+  "name": "demo-filesystem",
+  "protocol": "stdio",
+  "command": "npx"        // ← no "args", so there is no package to execute
+}
+```
+
+`npx`, `uvx`, `pipx` and `bunx` all take the package to run as their first
+argument. Without it they either print usage and exit or drop into an
+interactive prompt, and neither speaks MCP — so the connection fails on every
+attempt and the server stays permanently unhealthy.
+
+mcpproxy detects this *before* spawning the process, which is why the failure is
+instant rather than a timeout.
+
+## How to fix
+
+### Add the package to `args`
+
+```jsonc
+{
+  "name": "demo-filesystem",
+  "protocol": "stdio",
+  "command": "npx",
+  "args": ["-y", "@modelcontextprotocol/server-filesystem", "/path/to/serve"]
+}
+```
+
+`-y` (npm) skips the install prompt, which would otherwise hang the handshake.
+The `uvx` equivalent needs no flag:
+
+```jsonc
+{ "command": "uvx", "args": ["mcp-server-time"] }
+```
+
+### Inspect what is currently configured
+
+```bash
+mcpproxy upstream get <server> -o json
+```
+
+### Remove the entry if it is a leftover
+
+An entry that was added for a demo or a test and never completed is worth
+deleting rather than fixing — it keeps the tray badge red for a server nobody
+uses:
+
+```bash
+mcpproxy upstream remove <server>
+```
+
+### Apply the change
+
+The config file is hot-reloaded, so saving `~/.mcpproxy/mcp_config.json` is
+enough. To force a retry immediately:
+
+```bash
+mcpproxy upstream restart <server>
+```
+
+## Related
+
+- [`MCPX_STDIO_SPAWN_ENOENT`](MCPX_STDIO_SPAWN_ENOENT.md) — the command itself was not found
+- [`MCPX_CONFIG_PARSE_ERROR`](MCPX_CONFIG_PARSE_ERROR.md) — the config file is not valid JSON
--- a/docs/errors/MCPX_HTTP_LEGACY_SSE.md
+++ b/docs/errors/MCPX_HTTP_LEGACY_SSE.md
@@ -1,0 +1,63 @@
+---
+id: MCPX_HTTP_LEGACY_SSE
+title: MCPX_HTTP_LEGACY_SSE
+sidebar_label: LEGACY_SSE
+description: The endpoint rejected the streamable-HTTP initialize POST with a 4xx — the signature of a server that only speaks the older SSE transport.
+---
+
+# `MCPX_HTTP_LEGACY_SSE`
+
+**Severity:** error
+**Domain:** HTTP
+
+## What happened
+
+mcpproxy opened the connection with a streamable-HTTP `initialize` POST and the
+endpoint answered with a 4xx. That combination is the signature of a server
+implementing only the **older SSE transport**, which expects a `GET` to open an
+event stream and a separate POST endpoint for messages — so it rejects the
+streamable-HTTP handshake outright.
+
+This is a **transport mismatch, not an authentication or routing failure**. It
+is reported separately from `MCPX_HTTP_403` / `MCPX_HTTP_404` precisely because
+those send you looking for a credential or URL problem that does not exist.
+
+## How to fix
+
+### Set the transport explicitly
+
+```jsonc
+{
+  "name": "my-server",
+  "protocol": "sse",                       // was "http" or "streamable-http"
+  "url": "https://example.com/sse"
+}
+```
+
+### Check the URL path
+
+Legacy SSE servers usually publish the stream on a distinct path. If the
+endpoint documents `/sse` and the config points at `/mcp` (or the bare origin),
+correct the URL as well as the protocol.
+
+### Let mcpproxy pick
+
+Omitting `protocol`, or setting it to `auto`, makes mcpproxy negotiate — useful
+when the server supports both and you do not want to pin a choice:
+
+```jsonc
+{ "name": "my-server", "protocol": "auto", "url": "https://example.com/mcp" }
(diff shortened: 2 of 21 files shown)


Dumbris:
## Cross-model review + multi-agent verification — findings and fixes

Reviewed with **opencode / gpt-5.6-sol**, then independently verified by a 14-agent sweep (each of the four fixes verified by running real commands, every claimed defect handed to adversarial refuters). Three real defects found, all fixed in `5e773a98a` and `a17e09c4c`.

### The important one: `NSStatusBarButton` is FLIPPED

Verified by compiling against real AppKit rather than reasoning about it:

```
NSView().isFlipped          = false
NSButton().isFlipped        = true
NSStatusBarButton.isFlipped = true
```

So `9c740c0d4` — the first commit on this branch — shipped the badge in the icon's **top**-right corner, not the bottom-right its comment and commit message both claimed. In a flipped button `y = 2` is 2pt from the *top*, and `.maxYMargin` pins to the *top*. Rendered inside a real 36×22 status button, the dot landed at bitmap rows 4–15 of 44 (top-first).

`5e773a98a` had already corrected it, but **incidentally** — `TrayBadgeDotView` is a plain `NSView` and computes the dot in its own unflipped space. Nothing recorded that the design depends on that, and no test could tell the two apart: every badge test rendered the view standalone, where the flip never enters the picture.

Notably, the cross-model review did **not** catch this — I gave it the wrong premise (I told it the button was non-flipped). It correctly reasoned from what it was given. The multi-agent sweep caught it by compiling a probe instead of trusting either of us.

Fixed: `isFlipped` is now overridden to `false` explicitly and documented as load-bearing; the misleading comments are corrected with the trap written down; and `TrayBadgeInStatusButtonTests` renders the overlay inside a **real** `NSStatusBarButton`. Mutation-tested — flipping the overlay fails it with `badge is in the TOP half — the flipped-coordinate bug is back (rows 4...15 of 44)`, matching the shipped bug's measured position exactly.

### Anchor and stale-bounds (from the cross-model review, fixed in `5e773a98a`)

The 6×6 subview pinned with `[.minXMargin, .maxYMargin]` anchored to the **button's** corner, but the dot belongs to the **centred icon** — when the button shrinks by N the dot moves left by N while the icon's right edge moves left by only N/2. Compounded by `NSStatusItem` resizing asynchronously, so `button.bounds` read right after clearing `attributedTitle` was still the old, wider value.

The overlay now fills the button and derives the dot inside `draw(_:)` from live bounds, which removes both failure modes rather than patching either.

### `MCPX_HTTP_LEGACY_SSE` matches a string we don't own

The classifier comment, the test file and this PR's description all called it "mcpproxy's own error message". It isn't — it's mcp-go's exported `transport.ErrLegacySSEServer`. A routine dependency bump rewording it would silently regress the code to `MCPX_HTTP_404` with the whole suite green.

Corrected the attribution, added `errors.Is` as the primary check (the substring fallback stays, since the upstream layer stringifies the error and breaks the `errors.Is` chain), and added `TestLegacySSESentinelStillMatches` to pin the vendored wording so the bump fails a test instead of degrading in the field.

### Smaller consistency gaps

- `serversWithDiagnostic` was left on the narrower OAuth-only predicate when `isBadgeExempt` was introduced, breaking the "these filter identically" invariant at birth. Nothing consumes it today — which is exactly why the drift would have surfaced late. Aligned and pinned by a test.
- `isQuarantineReview`'s comment claimed `quarantined` and `admin_state` were two independent signals. They aren't: `health/calculator.go` derives the second from the first. The check is kept as belt-and-braces; the claim is corrected.
- Added the composition case nothing covered: servers are quarantined **on add**, so a newly added `npx`-with-no-args server is quarantined *and* broken. It asks for review rather than raising a red badge — deliberate, now pinned, along with the case that it *does* alarm once released from quarantine.

### Checked and cleared — not worth re-spending on

- No severity escalation from the new codes: `MCPX_UNKNOWN_UNCLASSIFIED` was already `SeverityError`, so badge colour is unchanged.
- Docs build is safe; all five cross-links in the new pages resolve and each target is already in `sidebars.js`. (Correction to the PR body above: `website/prepare-docs.sh` is a blanket copy — there is no allowlist at all.)
- No cross-platform gap: the severity badge is Swift-only; `internal/tray` has no badge concept.
- The `has no args` rule has exactly one non-test producer in the tree, so no false-positive surface.

### Known, deliberately out of scope

`frontend/src/views/ServerDetail.vue` renders the error diagnostic panel for a quarantined server with no exemption — the same one-payload contradiction, on a surface this PR doesn't touch. There's no aggregate error badge in the frontend, so there's no red-dot equivalent, but it's worth its own issue.

### Verification

`go build ./...` clean · `golangci-lint` (CI v2 config) **0 issues** · Go tests pass · Swift **1136 tests, 1 failure** — `AppLifecycleTests.testTheSharedJournalNeverWritesToTheRealInstanceRootUnderTests`, which asserts `~/.mcpproxy/tray-lifecycle.jsonl` doesn't exist and fails on any machine with a running tray. Pre-existing and environmental.

The corner dot still has not been seen by a human — Screen Recording isn't granted in this session — but it is now rendered and measured inside a real status button by the test suite, which is stronger than a one-off screenshot.

PR #1082, opened 2026-08-29
feat(settings): expose both serialization modes in the Web UI and tray

Dumbris:
## The gap

`tool_response_mode` (Spec 085) and `direct_tool_response_mode` (Spec 102) shipped with **four** ways to set them — the config file, an env var, a `serve` flag and the REST API — and **zero** UI surfaces. A tray-only operator could not reach `direct_tool_response_mode` at all, which is the entire point of the schema-deferred feature.

## What changed

Both fields now sit in **Settings → General**, directly under `routing_mode`, the knob they qualify — in the Web UI *and* the macOS tray.

Neither carries a restart badge. `internal/runtime/config_hotreload.go` appends both to `ChangedFields` **without** setting `RequiresRestart`, so they hot-apply — unlike `routing_mode` right above them, which genuinely needs one. Tests pin that asymmetry in both catalogues.

## The one new mechanism, and why it was needed

Both keys are `omitempty` on the Go side, where an absent value means `"full"`. Verified against a live instance — a fresh config returns **neither key**:

```
'routing_mode'              -> 'direct'
'tool_response_mode'        -> <ABSENT>
'direct_tool_response_mode' -> <ABSENT>
```

A native `<select>` (and a SwiftUI `Picker`) whose value matches no option renders an **empty box** — the default would have read as "unset". So `SettingField.defaultValue` declares the resolved default, and a normalize pass fills it at hydration in both clients.

Normalization is **display-only** and deliberately narrow:

- Applied to **both** the working copy and the last-saved snapshot, so an untouched field never counts as an unsaved change.
- **Never written back.** Saving one mode PATCHes only that key; the other stays absent from the config file.
- The **Raw JSON tab keeps showing server truth.** The Web UI already serialized the untouched response; the tray reused `original` for it, so this adds a separate `raw` copy rather than letting invented keys leak into that view.
- **Only the two opted-in fields declare a default**, and a test pins that exact list so the pass cannot quietly start inventing values for other unset keys.

## Docs defect found and fixed

`tool_response_mode` had **no published documentation**. The schema-deferred page sent readers to Search & Discovery for "its own `tool_response_mode` axis" — and that page never mentioned it. Now documented there, with the cross-reference repaired to a pinned anchor (`{#tool-response-mode}`, verified present in the built site), which also gives the new field a docs link that resolves.

## Verification

Against a **real instance** on an isolated port with a scratch data-dir and config (the user's own `~/.mcpproxy/` untouched; core torn down after):

- Both selects render **"Full — complete schemas (default)"** on a config containing neither key; **no** phantom dirty state, **no** unsaved-changes banner.
- Flipping the mode **through the Settings UI** saved as `Settings saved · direct_tool_response_mode` — *not* "restart required", matching the no-badge decision.
- `/mcp/all` **hot-switched with no restart**: `inputSchema` became exactly `{"type":"object"}`, descriptions gained their compact signature (`echo(message*:str)`), **tool count unchanged** (14 → 14, serialization-only), payload **8630 → 6523 bytes**.
- After reload: explicit `deferred` shows as Deferred, absent `tool_response_mode` shows as Full, neither dirty.
- Raw JSON tab read from the Monaco model: contains the real `direct_tool_response_mode`, does **not** contain the invented `tool_response_mode`.

Suites: frontend **973/973**; Swift **1149 tests, 1 failure** — `AppLifecycleTests.testTheSharedJournalNeverWritesToTheRealInstanceRootUnderTests`, which I confirmed fails identically on a clean tree with these changes stashed. `go build ./...` clean; docs site builds clean.

The frontend catalogue test was mutation-checked (dropping `defaultValue` fails it). One assertion I wrote was vacuous — comparing two identically-normalized objects passes unconditionally — and is rewritten to assert the divergence from the raw response, which is what gives the "normalize both copies" rule teeth.

## Not covered

The tray is verified at unit and compile level only, not visually — a live tray check needs the prod-bundle swap, which would disrupt the running tray and its managed core.


Dumbris:
## Cross-model review + 5-lens audit — 14 confirmed findings

Reviewed with **opencode / gpt-5.6-sol**, then a 5-lens audit fan-out (roadmap sync, spec sync, docs sync, adversarial code review, live end-to-end verification), each finding adversarially verified before action. **9 further findings were refuted and dropped.**

### The one that falsified this PR's own claim

The description above said a resolved default is "never written back". **That was wrong.** `SettingsSection` keeps a component-local `dirty` ref, and `dirtyKeys` is the UNION of that ref and the working/original comparison. Nothing keyed the section, so **Reload** replaced both snapshots while leaving the ref populated:

> pick **Compact** → click **Reload** → **Save** → PATCHes `tool_response_mode: "full"` into a config that never had the key.

Fixed by keying every section on a form epoch bumped per hydration. This also fixes the general case — the stale ref could strand *any* field, not just the new ones.

**Verified live** on a rebuilt binary: after Reload the abandoned edit is gone, Save is disabled, and the key is still absent from the config file on disk.

### Tests that could not fail

The earlier commit fixed a vacuous "both copies normalized" assertion in the **tray** test. The audit found I had left **the identical tautology in the Web UI twin**, plus a Raw-tab test that matched source text and so could not tell whether `cfg` was still the untouched response by the time it was serialized. Both now assert against an extracted `hydrateConfigState` helper — the Vue counterpart of the Swift `hydrate(from:)` seam. Mutation-checked both ways: normalizing one snapshot fails them; stripping the section keys fails the regression test.

### Also in the earlier commit (`cd0aee326`)

`DetectConfigChanges` compared `tool_response_mode` raw, so `"" → "full"` reported a change that never happened. The **direct** axis three lines below normalizes and documents exactly why — that landed in spec 102's own review and was never mirrored across. Exposing "full" as an explicit UI choice is what makes the round trip reachable.

### Roadmap / spec / docs sync

- **roadmap.yaml** — the `schema-deferred` epic still read `in_progress` with a note ending *"plan in progress"*; spec 102 shipped in #1063. Now `done` with a `pr:` ref, and the note's **"~88% smaller listing"** projection replaced with the measured 29.7% / 34.8% and the 38.9% ceiling. ROADMAP.md is **generated** — the edit is in the YAML, file regenerated via `scripts/gen-roadmap.py`.
- **spec.md / plan.md** — asserted SC-001's ≥70% as fact in **four more places**, and plan.md claimed *"Backend-only; no frontend, tray, or Swift changes"* and *"no tray state"*, both contradicted by this work. SC-001 is now marked **NOT MET** with the measurement, the ceiling, and the two options open to you. `tasks.md` was already honest — only two drift notes corrected (T083 claimed four new goldens, `git show` lists three; T073 named the wrong file).
- **docs** — *"`routing_mode` sits directly above it"* was wrong (`tool_response_mode` is between them). The settings-page restart list named `data_dir`, **which is not a field at all**, and omitted `routing_mode`.

### Filed separately rather than widened into this PR

- #1083 — `describe_tool` on the direct surface returns identifiers that are **not callable there** (`server:tool` and a `call_with` naming a tool the surface does not expose). Pre-existing spec-102 defect; breaks the discovery path, not the self-healing retry path.
- #1084 — invalid config value returns **500 instead of 400**; a malformed structured log line (`patcherrorconfiguration`); and `/api/v1/status` embedding a permanently stale copy of `upstream_stats`.

### Verification

Full frontend suite **978/978**. `internal/runtime` green under `-race` (154s). Swift **1153 tests, 1 failure** — the environmental `AppLifecycleTests` one, confirmed identical on a clean tree with changes stashed, and green in CI.

Live, on an isolated port with a scratch data-dir (user's `~/.mcpproxy/` untouched, core torn down): saving through the UI hot-switched `/mcp/all` with no restart — `inputSchema` → `{"type":"object"}`, signature appended, 14 tools unchanged, payload 8630 → 6523 bytes.

PR #747, opened 2026-06-22
feat(bench): token-reduction benchmark harness over frozen corpus (MCP-42)

Dumbris:
## What

First, fully-deterministic slice of the roadmap-#19 benchmark harness (MCP-42): the **token-reduction** numbers behind mcpproxy's "massive token savings" claim. In-repo under `bench/` (per board decision — no separate public repo).

Compares the static context-token cost of the three routing modes over a frozen tool corpus:

| Mode | Tools in context | Tokens | Savings |
|------|-----------------|--------|---------|
| `baseline` (all tools loaded) | 45 | 1730 | — |
| `retrieve_tools` (BM25 discovery) | 5 | 596 | **65.5%** |
| `code_execution` (orchestration) | 2 | 513 | **70.3%** |

These are a **conservative floor**: input schemas are excluded uniformly (the committed corpus has none), which understates the baseline; and savings scale with tool count (real deployments expose hundreds–thousands of tools).

## How

- Reuses the **Spec 065 frozen corpus** (`specs/065-evaluation-foundation/datasets/corpus_v1.tools.json`) as a versioned, non-drifting universe (CN-002).
- Tokenizer: `tiktoken cl100k_base` — already a repo dependency, reproducible, model-agnostic estimator. No new deps.
- Real proxy tool definitions captured verbatim from `internal/server/mcp.go` into `bench/proxy_tools_v1.json` (provenance recorded in-file).
- `go run ./bench/cmd/bench` → `report.json` + self-contained `dashboard.html` in `bench/results/` (gitignored; reports never committed per Spec 065 CN-003).
- Methodology, scoring rubric, dataset sources, known limitations, reviewer contact: `bench/README.md`.

## Tests

- TDD: `go test ./bench/` — deterministic tokenizer, per-mode tool exposure, real savings in (0,1), baseline monotonicity. Race-clean.
- `gofmt`, `go vet`, and **golangci-lint v2 (strict CI config)** all clean.

## Scoped but NOT in this PR (tracked as follow-ups)

These need decisions / other lanes, so they're deliberately deferred (see `bench/README.md`):
- **Live run** (docker-compose skeleton included): full schemas from `GET /api/v1/tools` for the exact headline number + Recall@k accuracy (reusing the Spec 065 retrieval golden set) + latency.
- **End-to-end task success with a pinned LLM** — needs a pinned model + LLM-call budget.
- **CI publish-on-release-tag → public dashboard** — Release/DevOps lane.

Related #MCP-42

PR #1137, opened 2026-08-31
docs(specs): 103-token-bench — measured savings, published results

Dumbris:
## Why

MCPProxy's headline promise is "massive token savings". Today that rests on a **static tool-surface** measurement — how many tokens the tool definitions cost before the agent does any work. That number is real, reproducible and CI-gated. It is also not what a user pays. **A mode that halves the menu but doubles failed attempts costs more, not less**, and nothing in the repo measures that.

Two facts make it urgent:

1. **Spec 102 projected ~70% and measured 29.7%** on the frozen 45-tool corpus (34.8% at 527 tools). SC-001 was restated per corpus shape rather than the design changed, and its gate concluded that *names, descriptions and annotations* — not schemas — dominate the payload, with 38.9% the arithmetic ceiling on that corpus. That conclusion comes from two frozen corpora and has never been checked against a real fleet.
2. **Retry rates are assumed, not observed.** Session-cost rows are computed from literature-derived per-arm defaults; the dashboard honestly labels them `estimated`. Every session-level savings claim inherits that assumption.

## Scope boundary — the main risk this spec manages

`bench/` is already a 74-file harness (PRs #747/#748/#851). The spec opens with a table naming what it already provides — encoding arms including spec 102's deferred rendering, token counting derived from the *live* tool builders, response cost, break-even, recall/nDCG, the provenance-badged v2 report, the frozen corpora — and states that **rebuilding any of it is a defect**.

Described in capability terms rather than file names, since a speckit `spec.md` must stay free of implementation detail.

What is genuinely missing, and is the scope here:

1. **Session replay** — nothing in `bench/` reads the activity log
2. **Measured** first-call success and retry rate, replacing the assumed defaults
3. **Tokens per *completed* task** as the headline, which needs a completion signal the static harness has no concept of
4. **Crossing the full three-axis mode matrix** (routing mode × discovery-surface serialization × direct-surface serialization), plus batching, stored scripts and validate-before-dispatch
5. **Payload decomposition** that confirms or corrects the spec-102 conclusion on real fleets

## Shape

4 user stories (P1: measure cost per completed task on replayed real sessions), **27 FRs, 10 success criteria, zero `[NEEDS CLARIFICATION]` markers**.

The criterion that matters most is **SC-006**: a mode that costs fewer surface tokens but completes fewer tasks must look *worse* in the headline metric. The benchmark is required to **detect a false saving rather than reward one** — which surface-only measurement structurally cannot do.

Also specified because it is the most dangerous silent failure: recorded responses truncated at capture time under-count tokens **in the project's favour**, so detection is mandatory and silent inclusion is forbidden (FR-002, SC-007).

## Assumptions resolved instead of clarification markers

- **Pinned model / spend** for agent-loop runs → plan-stage budget decision; the deterministic half is specified to stand alone without it.
- **A task-completion signal is derivable from recorded sessions** → assumed, with an explicit fallback. **This is the assumption most likely to be wrong** and the checklist names it as the first thing plan stage should validate.
- **Public suite left as a capability**, not named — the 2026-08 shortlist needs re-confirmation against the suites' current code (specifically whether they emit per-task token counts and can target a single proxy endpoint; FR-020/FR-022 are unsatisfiable if neither holds).

## Next

`/speckit.plan`. Cross-model spec review (opencode `gpt-5.6-sol`) is running against the tree to check for duplication with `bench/`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -158,7 +158,6 @@ tail -f ~/Library/Logs/mcpproxy/main.log  # main log (macOS; Linux: ~/.mcpproxy/
 - **Windows installer**: [docs/github-actions-windows-wix-research.md](docs/github-actions-windows-wix-research.md). **Prerelease** (`next` branch + `v*-rc.*` tags, opt-in, off stable channels): [docs/prerelease-builds.md](docs/prerelease-builds.md).
 
 ## Recent Changes
+- 103-token-bench: Go 1.25 (`bench/`) + existing only — tiktoken-go v0.1.8 (cl100k_base), the mcp-go transport already used by `bench/mcpcaller.go`. MCPMark is an external SHA-pinned tool invoked out of process, not a module dependency. **No new dependencies.**
 - 102-schema-deferred: Go 1.25.5 module toolchain (`go.mod`), backend-only + existing only — mark3labs/mcp-go v0.57.0 (tool surface), santhosh-tekuri/jsonschema/v6 (Spec-085 validator reuse), zap, stdlib. **No new dependencies.**
 - 101-tpa-db: Go 1.25.5 module toolchain (`go.mod`), backend-only + stdlib only for the new work — `crypto/ed25519`, `crypto/sha256`, `syscall` (no ed25519 usage in the tree before this); existing bbolt (state), zap, Cobra. **No new dependencies.**
-- 098-tools-preflight: Added Go 1.24 module toolchain (repo builds with local Go 1.25) + existing only — chi (httpapi), bbolt (storage), Bleve (index), zap (logging), Cobra (CLI), swaggo/swag v2 (contract regen). **No new dependencies.**
-- 097-stored-scripts: Added Go 1.25 (os.Root/Root.ReadFile available — R1) + stdlib only (os.Root). **No new dependencies.**
--- a/specs/103-token-bench/checklists/requirements.md
+++ b/specs/103-token-bench/checklists/requirements.md
@@ -1,0 +1,70 @@
+# Specification Quality Checklist: Token-efficiency benchmark — measured savings, published results
+
+**Purpose**: Validate specification completeness and quality before proceeding to planning
+**Created**: 2026-08-31
+**Feature**: [spec.md](../spec.md)
+
+## Content Quality
+
+- [x] No implementation details (languages, frameworks, APIs)
+- [x] Focused on user value and business needs
+- [x] Written for non-technical stakeholders
+- [x] All mandatory sections completed
+
+## Requirement Completeness
+
+- [x] No [NEEDS CLARIFICATION] markers remain
+- [x] Requirements are testable and unambiguous
+- [x] Success criteria are measurable
+- [x] Success criteria are technology-agnostic (no implementation details)
+- [x] All acceptance scenarios are defined
+- [x] Edge cases are identified
+- [x] Scope is clearly bounded
+- [x] Dependencies and assumptions identified
+
+## Feature Readiness
+
+- [x] All functional requirements have clear acceptance criteria
+- [x] User scenarios cover primary flows
+- [x] Feature meets measurable outcomes defined in Success Criteria
+- [x] No implementation details leak into specification
+
+## Notes
+
+Validation ran over three iterations; the following were fixed rather than waived.
+
+**Iteration 1 — implementation detail leaked into a spec that must stay declarative.**
+The draft named concrete files, flags, CLI commands, struct fields and PR numbers
+(`bench/session.go`, `armRetryRates`, `ResponseTruncated`, `-corpus-v2`, `#747`). All were
+removed from requirements and success criteria. This was the hardest constraint to hold,
+because the feature's defining risk — duplicating a harness that already exists — is
+naturally expressed in file names. It is instead expressed as the **Scope Boundary** table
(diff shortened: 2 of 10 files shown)


Dumbris:
## Cross-model review round 1 — spec restructured

Seven findings from opencode `gpt-5.6-sol`, all verified against the tree and all applied. **Two were Critical and changed the spec's premise**, so the PR description above is superseded on one key point.

### 1. Replay cannot answer what the spec asked of it

`ActivityRecord` carries tool calls, arguments, responses, status and work-session grouping — but **no prompt, no conversation, no model state and no completion oracle**. Replaying a recording under a different mode therefore cannot reveal how the agent would have *behaved*: which calls it would attempt, whether the first was right, whether it finished. The original US1 asked for exactly that and was unsatisfiable — as was the roadmap task's original framing ("activity-log sessions re-run under each mode combo").

**Fix**: a binding **Replay Boundary** section, and US1 split in two:
- **US1** — deterministic counterfactual *cost recomputation* over real workload shape. No model, byte-reproducible.
- **US2** — live agent loop under a pinned model, which now owns every success/completion figure.

FR-004 forbids inferring behaviour from a recording; Out of Scope names it as *forbidden*, not merely excluded.

### 2. Reproducibility and privacy were in conflict, and the privacy side was absent

The body-inclusive export deliberately returns **full unmasked values including detected secrets** — it is the compliance/incident-response surface, not a browsing one. The spec kept session contents out of published artefacts but said nothing about protecting them *during* replay, including transmission to model providers.

**Fix**: FR-006–FR-009 (sensitive by default; never transmitted to third parties; flagged records excluded or reduced; inputs live outside the repo with a documented deletion step) and SC-009. FR-030 now admits figures from private sessions are **not independently reproducible** and cannot be the sole support for a published claim.

### Also applied

3. **The headline metric did not penalise worse completion.** Tokens-per-completed-task can *improve* while completion falls, if tokens fall faster — the old SC-006 was not generally true. FR-018 requires completion rate at equal prominence, FR-019 adds a completion-regression threshold that overrides any token saving, SC-007 requires demonstrating it with a deliberately degraded mode.
4. SC-001 demanded "every mode combination" while FR-017 requires skipping invalid ones — scoped to every *valid* combination.
5. Core terms were undefined — added a binding Definitions block, which also separates corrective retries from infrastructure retries.
6. Three FRs restated existing harness capabilities — moved to **Inherited Constraints** (IC-001–IC-004) so they are satisfied, not rebuilt.
7. The capability axes were not a testable matrix — FR-016 states them as binary conditions over the routing modes where each is available.

Now **32 FRs, 12 success criteria, 4 inherited constraints**.

### Current-state claims independently confirmed by the reviewer

Nothing in the harness replays activity exports; retry rates are literature-derived defaults; first-call success is not measured; the three axes are not crossed; and the spec-102 figures (29.7% / 34.8% / 38.9% ceiling against a 70% projection) are accurate.

Dumbris:
## Cross-model review complete — 9 rounds, 38 findings, READY FOR TASKS

Reviewed with opencode `gpt-5.6-sol` against the tree. Every finding was independently verified before being applied; one reviewer claim (a glob miss) was rejected with evidence.

**Findings per round: 7 → 9 → 8 → 5 → 5 → 1 → 1 → 1 → 0.**

### The findings that changed the design, not the wording

**Replay cannot do what the spec originally asked.** `ActivityRecord` carries tool calls, arguments and responses — but no prompt, conversation, model state or completion oracle. So replaying a recording under a different mode cannot reveal how the agent would *behave*. US1 was split into deterministic cost recomputation (no model) and US2, a live agent loop that owns every success/completion figure. FR-004 now forbids inferring behaviour from a recording.

**A fleet input was missing entirely.** A menu cost is a property of the tool definitions, and the export carries no fleet snapshot — `ProxyToolsForMode` returns only built-in proxy tools, not the upstream fleet. The documented `-replay <jsonl>` invocation could compute *nothing* under `direct`. Replay now requires a frozen corpus or a live proxy.

**Bodies-off yields deltas, not absolutes.** Complete workload includes every consumed response; that text is absent. What survives is menu cost per cell plus the cross-mode delta between the two direct cells, whose identical call responses cancel.

**The privacy boundary contradicted the measurement design.** No content may leave `replaycorpus`, yet tokenization sat downstream. Resolved by tokenizing inside the loader and emitting counts only.

**Truncated `retrieve_tools` records cannot be identified.** `wasTruncated` is computed in the handler and only *logged* — the internal activity emitter has no truncation parameter. A loader would silently tokenize the stored full response and overstate what the agent paid. This forced a second backend change, with an explicit blanket-exclusion fallback.

### Errors this review caught in my own corrections

Three consecutive rounds found that a *fix* introduced the next round's defect — each an over-broad claim about what is measurable. The final correction is structural: a per-cell table separating **menu cost** from **complete workload cost**, which is the distinction the earlier drafts kept collapsing.

### Verified current-state claims

Nothing in `bench/` replays activity exports; retry rates are literature-derived defaults; first-call success is not measured; the three config axes are not crossed; spec 102's figures (29.7% / 34.8% / 38.9% ceiling vs a 70% projection) are accurate; the export path never masks; `has_sensitive_data` is derived from asynchronously-added metadata; code-execution sub-calls emit zero byte counts; `bench.yml` runs only on `v*` tags so an SC-011 gate there could never block a PR.

PR #1139, opened 2026-08-31
docs(specs): 103-token-bench tasks — 55 tasks, US1 shippable alone

Dumbris:
## What

`/speckit.tasks` output for spec 103, following the spec and plan merged in #1137.

**55 tasks across 8 phases** — Setup (3), Foundational (11), US1 (14), US2 (13), US4 (4), US3 (5), US5 (1), Polish (4). 22 parallelizable.

**Tests are included, not optional**: constitution Principle V requires them before implementation, so each production change is preceded by its failing test.

## MVP

**Phase 1 + Phase 2 + US1.** That delivers real-workload menu costs and the `direct_full` vs `direct_deferred` cross-mode delta with **zero model spend**, replacing frozen-corpus numbers with numbers from real fleets. It does not depend on the pinned-model decision that gates US2.

## Sequencing that came out of the review

**Phase 1 fixes two reproducibility defects before anything depends on them.** The tokenizer cache must be *populated*, not merely named — setting `TIKTOKEN_CACHE_DIR` alone still downloads on first run, which defeats SC-004 offline. And the SC-011 no-committed-reports gate **cannot live in `bench.yml`**, which triggers only on `v*` tags and `workflow_dispatch` and so could never block a PR.

**Phase 2 carries both production changes**, each preceded by a failing test:
- export byte fields (`internal/contracts/activity.go`, `internal/httpapi/activity.go`)
- truncation-flag propagation (`internal/server/mcp.go`) — without it a truncated `retrieve_tools` record cannot be identified and the loader would silently overstate agent cost

Plus the additive report blocks (no `report_version` bump) and the 5-cell mode matrix.

**Dependencies**: US1 and US2 are independent once Phase 2 lands. US4 needs US1's fleet plumbing. US3 needs both P1 stories to have produced figures. US5 needs US3 — publishing before reproducibility is the exact failure this feature exists to avoid.

## Five standing warnings

Restated at the end of the file, each traceable to a defect the 9-round review found:

1. Never let a truncated record contribute silently — it understates cost *in the project's favour*.
2. Never sum across accounting sources; withhold with a stated reason.
3. Never publish a percentage without its fleet shape.
4. Never present a bodies-off figure as an absolute complete-workload cost.
5. A recording-only run is an **error**, not a degraded run — without a fleet input there is no menu to cost.

## Note

`ROADMAP.md` regenerated: the new `tasks.md` adds `103-token-bench` to the per-spec progress table at 0/55.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -910,4 +910,4 @@ Legend: `shipped` ≥95% checked · `in-flight` 1–94% · `drafted` 0% · `—`
 | [100-prompt-rugpull-baseline](./specs/100-prompt-rugpull-baseline/) | — | — |
 | [101-tpa-db](./specs/101-tpa-db/) | — | — |
 | [102-schema-deferred](./specs/102-schema-deferred/) | `shipped` | 89/89 (100%) |
-| [103-token-bench](./specs/103-token-bench/) | `drafted` | 0/59 (0%) |
+| [103-token-bench](./specs/103-token-bench/) | `drafted` | 0/64 (0%) |
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -910,4 +910,4 @@ Legend: `shipped` ≥95% checked · `in-flight` 1–94% · `drafted` 0% · `—`
 | [100-prompt-rugpull-baseline](./specs/100-prompt-rugpull-baseline/) | — | — |
 | [101-tpa-db](./specs/101-tpa-db/) | — | — |
 | [102-schema-deferred](./specs/102-schema-deferred/) | `shipped` | 89/89 (100%) |
-| [103-token-bench](./specs/103-token-bench/) | `drafted` | 0/55 (0%) |
+| [103-token-bench](./specs/103-token-bench/) | `drafted` | 0/59 (0%) |
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -910,3 +910,4 @@ Legend: `shipped` ≥95% checked · `in-flight` 1–94% · `drafted` 0% · `—`
 | [100-prompt-rugpull-baseline](./specs/100-prompt-rugpull-baseline/) | — | — |
 | [101-tpa-db](./specs/101-tpa-db/) | — | — |
 | [102-schema-deferred](./specs/102-schema-deferred/) | `shipped` | 89/89 (100%) |
+| [103-token-bench](./specs/103-token-bench/) | `drafted` | 0/55 (0%) |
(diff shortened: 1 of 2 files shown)


Dumbris:
## Task review complete — 4 rounds, 15 findings, READY TO IMPLEMENT

Reviewed with opencode `gpt-5.6-sol`. **Findings per round: 10 → 4 → 1 → 0.**

**55 → 64 tasks.** The growth is the point: four of round 1's findings were **coverage gaps** — work the plan requires that no task covered.

### Coverage gaps found

| Gap | Why it mattered |
|---|---|
| **FR-020 baseline arm** never implemented | Nothing ran the same agent over the same tasks with all tools loaded directly, *bypassing mcpproxy* — the denominator every published percentage is measured against. The deterministic `baseline` renderer used by the offline arms is a different thing and doesn't substitute. |
| **SC-010 had no execution task** | Patching the suite, configuring the fleet and choosing tasks were all present; actually *running* it across two cells was not. |
| **FR-023 cost-versus-outcome view** | Emitting a data block isn't rendering cost against completion — which is the whole point of not rewarding a false saving. |
| **FR-016 capability conditions** | Batching, stored scripts and validate-before-dispatch were neither tested nor implemented. |

### A regression the review caught in my own fix

Round 1's fix rewrote US2 to add the baseline arm and, in doing so, **silently dropped the FR-013 task** — the one making measured retry rates supersede the literature-derived defaults.

Root cause: tasks cross-referenced each other by ID (`Same file as T032`), and renumbering invalidated those notes with no signal. Beyond restoring the task, **every intra-file task-ID reference is now a descriptive phrase**, so renumbering can't silently break a dependency note again.

### Ordering and hygiene

- **TDD ordering was inverted in five places** across rounds — both backend changes, the report/schema tests, the US3 report behaviours, and the FR-004 pair. All now lead with a failing test, as constitution Principle V requires.
- **Four `[P]` markers sat on tasks sharing a file** (`bench/replay_test.go`, `bench/agentloop_test.go`, `bench/payloaddecomp_test.go`, `bench/README.md`). `[P]` means *different files*; the parallel-opportunities section had also claimed six test tasks were "distinct test files" when three were not.
- **FR-004 had no implementation or test.** Replay output must be labelled a *counterfactual over recorded traffic* — the Replay Boundary is the spec's most load-bearing constraint, and nothing enforced it.

Final: 64 tasks, 8 phases, 8 parallelizable, every FR and SC covered, no brittle ID references.

PR #1156, opened 2026-09-01
fix(docker): pick a git-capable isolation image, and classify container-toolchain failures

### LINKED ISSUE #1143: Default uvx isolation image is -slim and has no git, so every git+https:// MCP server fails under isolation
## Summary

The default Docker image for `uvx` / `python` / `pip` / `pipx` is a `-slim` variant that ships **no `git`**. Any MCP server installed from a git URL therefore can never start under Docker isolation.

## Evidence

[`internal/config/config.go:1535-1539`](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/main/internal/config/config.go#L1535-L1539) defaults these runtimes to `ghcr.io/astral-sh/uv:python3.13-bookworm-slim`.

```
$ docker run --rm --entrypoint sh ghcr.io/astral-sh/uv:python3.13-bookworm-slim -c "command -v git || echo NO GIT"
NO GIT

$ docker run --rm --entrypoint sh ghcr.io/astral-sh/uv:python3.13-bookworm -c "git --version"
git version 2.39.5
```

## Real-world failure

`gcore-mcp-server`, configured as the vendor documents it:

```json
{ "command": "uvx",
  "args": ["--from", "gcore-mcp-server@git+https://github.com/G-Core/gcore-mcp-server.git", "gcore-mcp-server"] }
```

Container stderr:

```
Updating https://github.com/G-Core/gcore-mcp-server.git (HEAD)
× Failed to download and build `gcore-mcp-server @ git+https://github.com/G-Core/gcore-mcp-server.git@394e9ca`
├─▶ Git operation failed
╰─▶ Git executable not found. Ensure that Git is installed and available.
```

This is not a niche shape. A large share of MCP servers ship only as a git URL with no PyPI release (`gcore-mcp-server` is a 404 on PyPI), so for those servers Docker isolation is unusable out of the box. The user's only escape is to discover the cause from container stderr and hand-write a per-server image override.

## Suggested fix

Don't bump the global default to non-slim — that costs every user ~700MB for a case most don't hit.

[`GetDockerImage`](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/main/internal/upstream/core/isolation.go#L309) already receives the full `*config.ServerConfig`, so it can inspect args:

- if no per-server `isolation.image` override is set, **and** the runtime is a Python package runner, **and** any arg contains `git+` → resolve to a git-capable image (e.g. a new `default_images["uvx-git"]` key, defaulting to `ghcr.io/astral-sh/uv:python3.13-bookworm`).

Keeping it a named `default_images` key rather than a hardcoded string means users can still retarget it.

Related: #1144 (the resulting error is unclassified and tells the user to file a bug).

### LINKED ISSUE #1144: Container-toolchain failures classify as MCPX_UNKNOWN_UNCLASSIFIED and tell the user to file a bug
## Summary

Toolchain failures reported on the isolated container's **stderr** are not classified, so they fall through to `MCPX_UNKNOWN_UNCLASSIFIED`, whose user message is *"mcpproxy could not classify this failure. Please file a bug report so we can add a specific code."* — shown for a problem the user can fix in one line of their own config.

## Evidence

Feeding the real error (from #1143) through the classifier with the docker hint:

```go
Classify(err, ClassifierHints{DockerIsolated: true})
// → "MCPX_UNKNOWN_UNCLASSIFIED"
```

where `err` is the actual runtime error:

```
failed to connect: stdio transport (command="/usr/local/bin/docker", docker_isolation=true):
server process exited before completing the MCP initialize handshake; recent stderr:
  | ├─▶ Git operation failed
  | ╰─▶ Git executable not found. Ensure that Git is installed and available.
```

[`classifyDocker`](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/main/internal/diagnostics/classifier.go#L106) has no git pattern, and the message doesn't incidentally match the image-pull, permission, or OCI arms either.

## Why it matters

`classifier_selfmessages_test.go` exists specifically to enforce *"mcpproxy's own error message must never ask the user to file a bug"*. The same principle applies here: this isn't mcpproxy's message, but it is just as much a user-fixable configuration problem, and the "file a bug report" CTA is the worst possible response to it.

The general gap: **container-toolchain stderr is a different failure surface from mcpproxy's own errors**, and the classifier was built mostly for the latter. Anything the isolated image happens to lack lands in UNKNOWN — git today, `uvx` missing from `python:3.11` previously (the case `MCPX_DOCKER_EXEC_NOT_FOUND` was added for).

## Suggested fix

Add a `MCPX_DOCKER_MISSING_TOOLCHAIN` code (or extend `DockerExecNotFound`) matching, under the `DockerIsolated` hint:

- `git executable not found`
- `git operation failed`
- the generic `<tool>: not found` / `command not found` shapes seen in container stderr

Remediation text should name the concrete fix: set a per-server `isolation.image` that contains the missing tool, with the `-bookworm` example spelled out. A `docs/errors/MCPX_DOCKER_MISSING_TOOLCHAIN.md` page alongside the existing `MCPX_DOCKER_EXEC_NOT_FOUND.md` would fit the established pattern.

Worth pairing with a test asserting that a container-stderr sample never classifies as UNKNOWN, mirroring the self-messages test.

Dumbris:
Fixes #1143. Fixes #1144.

## #1143 — the default uvx image cannot install most MCP servers

`default_images` mapped `uvx`/`python`/`pip`/`pipx` to `ghcr.io/astral-sh/uv:python3.13-bookworm-slim`, which has **no git**. A large share of MCP servers ship only as a git URL with no PyPI release, so under Docker isolation they could never install:

```
├─▶ Git operation failed
╰─▶ Git executable not found. Ensure that Git is installed and available.
```

`GetDockerImage` now resolves a git-capable image when a server's args carry `git+` and no per-server `isolation.image` override is set, via a named `default_images["uvx-git"]` key operators can retarget. The global default stays slim — nobody pays ~700MB for a case they don't hit.

**An operator's own image map wins.** Review caught that the first attempt hardcoded a public `ghcr.io` URL that beat a mirrored/air-gapped operator's configured `default_images.uvx`, because the load-time merge seeds the key in *every* install so the "key absent" fallback never ran. Resolution now decides on **values**, not key presence: the operator's `uvx-git` wins outright; an explicitly-empty key opts out; a still-shipped built-in next to a *retargeted* runtime image stands down to the operator's image and logs a Warn naming the key to set. Pinned by a test that models the real load path — config file → `LoadFromFile` → merge → resolve — rather than hand-building the merged map.

## #1144 — the resulting error told users to file a bug

The failure above classified as `MCPX_UNKNOWN_UNCLASSIFIED`, whose message is *"Please file a bug report so we can add a specific code."* — for a one-line config problem. New `MCPX_DOCKER_MISSING_TOOLCHAIN` covers container-toolchain failures, with a docs page following the existing `MCPX_DOCKER_EXEC_NOT_FOUND` pattern.

## The hard part: not hijacking correct diagnoses

Container stderr is attacker-adjacent free text, and four separate review rounds found the new arms stealing classifications they shouldn't. Final behaviour, probed with the **realistic wrapper shape** mcpproxy actually produces (`command="/usr/local/bin/docker", docker_isolation=true`):

| stderr | code |
|---|---|
| `Git executable not found` (the real failure) | `MCPX_DOCKER_MISSING_TOOLCHAIN` |
| `git@github.com: Permission denied (publickey)` | `MCPX_STDIO_EXIT_BEFORE_INITIALIZE` |
| `Cannot connect to the Docker daemon` (on exit) | `MCPX_DOCKER_DAEMON_DOWN` |
| same wording in a *live* container's stderr (timeout) | `MCPX_STDIO_HANDSHAKE_TIMEOUT` |
| `/bin/sh: 1: source: not found` | `MCPX_STDIO_EXIT_BEFORE_INITIALIZE` |
| `docker: permission denied … daemon socket` | `MCPX_DOCKER_NO_PERMISSION` |

Three of these were bugs found during review, each the same shape — a guard applied to one arm but not its sibling:

- **git SSH auth reported as "Docker permission denied."** `classifyDocker`'s arm is `contains("permission denied") && contains("docker")`, and *mcpproxy's own error wrapper injects the docker command into the string* — so under isolation that arm was near-unfalsifiable and swallowed any permission-denied text from child stderr.
- **A shell builtin (`source: not found`) read as a missing image toolchain**, sending users to rebuild an image over a `#!/bin/sh` line. Now requires that *every* missing name be a non-builtin.
- **Daemon wording in a live container's log reported as `DAEMON_DOWN`.** A handshake timeout means the container is alive, so a layer failure cannot also be true — the wording came from the server's own log. A genuine daemon refusal is unaffected: it surfaces on the exit path.

## Verification

Four fix→re-review rounds (3-lens adversarial). The final round's "clean" verdict was **wrong** — its probes omitted the wrapper prefix, which is exactly how the git-auth misdiagnosis survived. Caught by hand-probing with the realistic shape; tests now pin that shape so it can't recur.

Gates: `go build ./...`, `-tags server`, `go test ./internal/... -race`, golangci-lint v2 → 0 issues.


Dumbris:
## Correction: the air-gap claim in the PR description is not yet true

An independent `opencode` (gpt-5.6-sol) cross-review found five defects, and I reproduced the two High ones against this branch. One of them contradicts a claim I made above, so flagging it before anyone reads the description as fact.

**The description says an operator's mirrored/air-gapped configuration wins. That holds only if they set `default_images`.** An operator who sets *only* `docker_isolation.registry` still pulls public GHCR for a git-dependency server:

```go
const DefaultGitCapableImage = "ghcr.io/astral-sh/uv:python3.13-bookworm"  // fully qualified
```

and `buildFullImageName` (isolation.go:433) returns any image that already contains a registry unchanged — so the configured registry is ignored. Earlier rounds fixed the `default_images` path and never covered the bare-`registry` path.

**Second High: the arm-exemption pattern again.** `DockerCLINotFound` and `DockerOCIRuntime` are still exempt from the handshake-timeout guard, so a live container's own log hijacks the diagnosis. Probed with the realistic wrapper:

| child stderr during a handshake timeout | got | want |
|---|---|---|
| `app: docker: command not found` | `MCPX_DOCKER_CLI_NOT_FOUND` | `MCPX_STDIO_HANDSHAKE_TIMEOUT` |
| `app: OCI runtime warning ignored` | `MCPX_DOCKER_OCI_RUNTIME` | `MCPX_STDIO_HANDSHAKE_TIMEOUT` |
| `app: exec format error while probing plugin` | `MCPX_DOCKER_OCI_RUNTIME` | `MCPX_STDIO_HANDSHAKE_TIMEOUT` |

The standing rationale for exempting these arms — "it means no container ever started" — is exactly as false for them as it was for the daemon arm that round 4 fixed: if the handshake *timed out*, a container is alive.

Three Mediums also stand: benign exit stderr stealing `STDIOExitBeforeInitialize`; `Git operation failed` defaulting to "missing toolchain" unless the cause is on a finite exclusion list (so `fatal: could not read from remote repository` claims git is absent even though git demonstrably ran); and an operator who explicitly sets `uvx-git` to the shipped value having that choice read as unset, because intent is inferred by value inequality.

**Do not merge this PR until the follow-up lands.** I will update the description once the claims in it are true.

Dumbris:
## Resolved — the do-not-merge hold is lifted

All five opencode findings are fixed, plus two more the follow-up review found. Verified by running the real code with the realistic wrapper shape:

| child stderr | code |
|---|---|
| `Git executable not found` (the original failure) | `MCPX_DOCKER_MISSING_TOOLCHAIN` |
| `git@github.com: Permission denied (publickey)` | `MCPX_STDIO_EXIT_BEFORE_INITIALIZE` |
| `Cannot connect to the Docker daemon` (on exit) | `MCPX_DOCKER_DAEMON_DOWN` |
| same wording from a *live* container (timeout) | `MCPX_STDIO_HANDSHAKE_TIMEOUT` |
| `/bin/sh: 1: source: not found` | `MCPX_STDIO_EXIT_BEFORE_INITIALIZE` |
| `docker: permission denied … daemon socket` | `MCPX_DOCKER_NO_PERMISSION` |

**The air-gap claim in the description is now true.** `DefaultGitCapableImage` was split into registry + repo and is qualified with `docker_isolation.registry`, so an operator who sets only `registry` gets `<registry>/astral-sh/uv:python3.13-bookworm`. Pinned by a test that runs the real `LoadFromFile` → merge → resolve path with no `default_images` at all.

Two structural changes rather than more arm-patching:
- `dockerDaemonEvidence` — the docker-permission arm now requires text only the daemon or a socket dial can produce. Previously it keyed on `contains("permission denied") && contains("docker")`, and **mcpproxy's own wrapper injects `command="/usr/local/bin/docker"` into every isolated error**, making that arm near-unfalsifiable.
- `uvx-git` is no longer seeded into `default_images`, so its *presence* is the operator's intent. Value comparison could not express "explicitly chose the shipped default", because `SaveConfig` marshals the merged map back into the operator's file — anything mcpproxy seeds reappears there indistinguishable from something they typed.

`git operation failed` also no longer defaults to "missing toolchain": git is claimed absent only when it printed no diagnostic of its own, so `fatal: could not read from remote repository` is no longer reported as git being missing.

### Merge-conflict resolution worth a look

main's `internal/diagnostics/hints` extraction supersedes this branch's inline hint building, so the conflict took main's side; `DockerArgs` was added there because the #1143 remediation reads it.

**#1155's `TestParkableCode_ChildStderrNeverProvesPermanence` needed updating**, and I want that visible rather than buried in a merge commit. It pinned the *old* classifications to isolate its permanence gate — e.g. a child printing `ENOENT` for its own data file was asserted to classify as `MCPX_STDIO_SPAWN_ENOENT`. This PR fixes exactly that misclassification, so the expected codes move. The gate's actual invariant (`parkable == false` for stderr-only evidence) is unchanged and still asserted, and I added two rows whose stderr still reaches a **permanent** code so the gate is never left incidentally untested.

PR #1159, opened 2026-09-01
docs(test): record why TestFastPathDoesNotOvertakeAQueue must park rather than release

Dumbris:
# Pull Request

## Description

> **Superseded in substance by main.** This PR opened as a de-flake of `TestFastPathDoesNotOvertakeAQueue` (`internal/upstream/limiter`). While it was open, main landed its own fix for the same flake, with the same diagnosis and a stronger implementation. After merging main in, **the net diff is 8 lines of comment.** Retained only if that comment is judged worth keeping; otherwise close it — the flake is fixed either way.

The flake blocked at least three unrelated PRs — #1037 (telemetry-only diff) and #1153 twice in a row (a `bench/`-only diff) — in both *Unit Tests* and *End-to-End Tests* on `ubuntu-latest`.

**The previously recorded root cause was wrong.** It was characterised as "the assertion demands strict FIFO from a fast path that only *usually* loses the race", i.e. a real ordering bug in `limiter.go`. It is not, and `limiter.go` is unchanged here.

Strict FIFO **is** a genuine invariant the limiter promises (package doc, FR-004) and it holds **deterministically** in this test. By the time the test calls `held()`, both callers are already parked in the queue — `Queued == 2` is asserted beforehand — so the fast path is not in play at all. Grant order is then fixed under one mutex: `grantLocked` grants the *front* waiter only (`Max: 1`), and the second waiter can only be granted from inside the **first waiter's `release()`**.

The flake was in the test harness, which released before reporting:

```go
if aerr == nil { release() }   // <- this is what wakes the second caller
first <- aerr                  // <- reported only afterwards
```

`release()` unblocks the second goroutine, so both channel sends become concurrently ready, and a `select` over two ready cases picks **uniformly at random** (Go spec). The test read a scheduling artifact as a FIFO violation. On an idle machine the first goroutine almost always sends before the second is even scheduled — which is why it never reproduced locally.

## What changed

Main's fix (`startParkedCall`) is the stronger one and is kept: each admitted call **parks** in the slot it wins instead of releasing, so with a cap of 1 only one of the two can be past admission at a time. The concurrent report becomes *impossible* rather than merely unlikely, and occupancy is asserted between the two grants. This branch's `atomic` sequence stamping and its `awaitAdmission` helper are dropped as superseded.

What survives from this branch is the doc comment recording **why** the earlier version failed, so the release-before-report trap is documented next to the helper that avoids it. That is the entire remaining diff.

## Testing

- [x] All existing tests pass

Verified **on the merge result**, so these cover main's implementation as shipped:

| Check | Result |
|---|---|
| `TestFastPathDoesNotOvertakeAQueue`, 12 CPU load-generators, `-count=800` | ok |
| Full package `-race -count=5` | ok |
| Mutant: `grantLocked` grants from `Back()` (LIFO) | caught |
| Mutant: fast path drops the `waiters.Len() == 0` guard | caught |
| `go build ./...` | ok |

For reference, before either fix: the pre-fix test **failed** under that identical 12-load-generator `-count=800` stress.

## Notes for the reviewer

- The generalisable lesson, and the reason the comment is proposed for keeping: **a `select` over two result channels can never prove ordering.** When goroutine A's completion is what unblocks goroutine B, both sends land in the same instant. Worth suspecting the harness first for any Go test claiming "X happened before Y" by racing two channels.
- Repro technique, since `-count=N` under load is unreliable: copy the test verbatim into a scratch file in the same package and insert `for i := 0; i < 50; i++ { runtime.Gosched() }` at the suspected preemption point. That failed 20/20 instantly where `-count=600` under load had not reproduced at all.


Dumbris:
Closing: main landed its own fix for this flake (`startParkedCall`) while this PR was open — same diagnosis, stronger implementation. After merging main in, the net diff was 8 lines of comment, which isn't worth a PR on its own. The flake is fixed on main either way.

Issue #1161, opened 2026-09-01
Server edition: every authenticated user can read shared servers' credentials via listServers/getServer


Dumbris:
## Summary

In the **server edition**, `listServers` and `getServer` return `ServerResponse`, which embeds the raw `*config.ServerConfig`, to **every authenticated user** — including the admin-configured *shared* servers. Nothing redacts them, so every user of a multi-user deployment can read the shared servers' credentials.

`internal/serveredition/api/user_handlers.go` — `listServers` (~line 208) and `getServer` (~line 326).

## What is exposed

Everything `config.ServerConfig` carries for a shared server sourced from `h.sharedServers`:

- `env` values
- `headers` values
- URL query credentials (`?token=…`) and userinfo (`user:pass@`)
- `oauth.client_secret`
- `auth_broker.client_secret` (`internal/config/auth_broker.go:44`)

## Why it matters here

The personal edition's equivalent doors are being fixed in [#1157](https://github.com/smart-mcp-proxy/mcpproxy-go/pull/1157) (MCP read surfaces, `GET /api/v1/servers`, and the `/events` SSE payload all now go through one redaction rule). This door is the same class of leak with a **worse trust model**: it is not "anyone who can reach an unauthenticated local port", it is "any authenticated tenant reads the admin's shared-server credentials".

Shared servers are precisely the ones configured once by an admin with real credentials and exposed to many users, so they are the worst case for this leak.

## Why it was not fixed in #1157

Deliberately out of scope, and flagging rather than silently expanding that PR:

- different edition and trust model (spec 024 multi-user) than the issue #1157 addresses;
- it was not among the reviewed findings on that branch;
- the clean fix wants the shared redaction walker that #1157 is landing in `internal/oauth`, so this is better done *after* it, on top of a stable API.

## Suggested fix

Once #1157 lands, route these handlers through `oauth.RedactServerSecretFields` (or the redacted server view) rather than embedding `*config.ServerConfig`. Embedding a raw config struct in a response is the recurring root cause across all of these — #1157's own `upstream_servers list` had two such sites for the isolation config.

A response type that cannot hold a raw `*config.ServerConfig` at all would prevent the next instance structurally.

Found during the #1157 review rounds.

Issue #1148, opened 2026-08-31
quarantine_security list_quarantined returns env/headers/oauth secrets in plaintext, unauthenticated


Dumbris:
## Summary

`quarantine_security operation=list_quarantined` serialises raw `config.ServerConfig` values, returning every quarantined server's `env`, `headers`, and `oauth.client_secret` **in plaintext** — over the `/mcp` endpoint, which is **unauthenticated by default** (`require_mcp_auth: false`).

## Reproduction

Verified against `main` (0750fcd) on a scratch instance. Every request below was sent to `/mcp` with **no `X-API-Key` header and no credential of any kind**.

Server configured with a secret env var and a secret header:

```
env: {'QA_SECRET_TOKEN': 'sk-super-secret-value-12345'}
headers: {'X-AuthToken': 'hdr-secret-98765'}
```

Then, unauthenticated:

```json
{"method":"tools/call","params":{"name":"quarantine_security",
 "arguments":{"operation":"list_quarantined"}}}
```

Response:

```
LEAKS env secret: True
LEAKS header secret: True
...,"env":{"QA_SECRET_TOKEN":"sk-super-secret-value-12345"},
    "headers":{"X-AuthToken":"hdr-secret-98765"},...
```

## Why it is reachable unauthenticated

[`internal/server/mcp.go:4136`](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/main/internal/server/mcp.go#L4136) marshals the storage result directly:

```go
servers, err := p.storage.ListQuarantinedUpstreamServers()   // []*config.ServerConfig
...
jsonResult, err := json.Marshal(map[string]interface{}{"servers": servers, ...})
```

`config.ServerConfig` carries `Env`, `Headers` and `OAuth` — none are masked on this path.

The agent-token gate is:

```go
if authCtx := auth.AuthContextFromContext(ctx); authCtx != nil && !authCtx.IsAdmin() {
	// reject
}
```

A **nil** auth context — an unauthenticated `/mcp` session — passes straight through. The gate blocks scoped agent tokens but not the absence of a token. `/mcp` is documented as intentionally unprotected for client compatibility, so on a default install this is exposed to anything that can reach the listener, including a prompt-injected agent already connected to the proxy.

## Impact

[...]

Issue #1142, opened 2026-08-31
isolation.enabled: nil (inherit global) is reported as false, so isolated servers display as unisolated


Dumbris:
> **Corrected 2026-08-31.** This issue originally claimed that `upstream_servers patch` *disabled* isolation. That was wrong — verified below. The persisted config is correct; the **reporting** is not. Title and body rewritten to the actual defect.

## Summary

`IsolationConfig.IsEnabled()` returns `false` when `Enabled` is `nil`, but `nil` means **"inherit global"** — as the field's own comment says, and as the spawn path implements. Every read surface (REST API, CLI, Web UI, tray) therefore reports Docker isolation as **disabled** for servers that are actually **running isolated**.

## The contradiction

[`internal/config/config.go:801`](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/main/internal/config/config.go#L801) — the field is explicitly tri-state:

```go
Enabled *bool `json:"enabled,omitempty"` // nil = inherit global
```

[`internal/config/config.go:833`](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/main/internal/config/config.go#L833) — but the accessor collapses inherit to off:

```go
func (ic *IsolationConfig) IsEnabled() bool {
	if ic == nil || ic.Enabled == nil {
		return false   // <-- "inherit global" reported as "disabled"
	}
	return *ic.Enabled
}
```

[`internal/upstream/core/isolation.go:219`](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/main/internal/upstream/core/isolation.go#L219) — meanwhile the **spawn path** gets it right, downgrading only on an explicit `false`:

```go
// (3) Global isolation active: honor a per-server bool opt-out.
if serverConfig.Isolation != nil &&
	serverConfig.Isolation.Enabled != nil && !*serverConfig.Isolation.Enabled {
	return config.IsolationModeNone
}
return globalMode
```

So with global isolation on and a per-server `isolation` object that has an image but no `enabled`, the server **is** containerized while every UI says isolation is off.

## Verification

Isolation object built by unmarshalling `{"image": "…bookworm"}` and merged onto a server with no prior isolation:

```
after unmarshal: Enabled=<nil> (nil=true) Image="ghcr.io/astral-sh/uv:python3.13-bookworm"
after merge:     Enabled=<nil> (nil=true) Image="ghcr.io/astral-sh/uv:python3.13-bookworm"
[...]

PR #1164, opened 2026-09-01
fix(security): route the remaining server-payload doors through one redaction rule

Dumbris:
Follow-up to #1157. Those review rounds continued after the branch was merged via #1162, and **rounds 8 and 9 never reached main**. This carries them, rebased onto current main.

## The one that matters: a mask that overwrites the credential it hides

Verified on a clean checkout of `main` (`3e6c4202b`) by running the real code:

```
masked = https://host/mcp?access_token=%E2%80%A2%E2%80%A2%E2%80%A2%E2%80%A2UE+%2825+chars%29
ContainsMaskMarker = false
```

`net/url` re-serialises the URL after component-wise masking, so the mask comes back **percent-encoded** — and `ContainsMaskMarker`, which *is* the write-side net (`UnmaskLiveURL`'s residual check and `CheckServerWriteMasks`), does not recognise its own output. An echoed masked URL that the key-bound revert cannot bind is therefore accepted and **persisted over the live credential**.

That turns a disclosure bug into silent credential destruction — the failure mode #1142 was about — and it is live on main today. Same probe on this branch returns `true`.

Fixed by deriving an escaped marker (`url.QueryEscape`) for every marker, plus registry entries for the two URL renderings.

## Read doors that publish credentials in the clear on main

| door | leaked |
|---|---|
| `internal/server/server.go:1195` `GetUpstreamStats` | `url`, `last_error` raw on `GET /api/v1/status` and every SSE status event — and this is the implementation that actually runs; `Manager.GetStats` is only its fallback |
| `internal/httpapi/server.go` `handleGetConfig` | the **entire** raw `config.Config`: every server's env, headers, `oauth.client_secret`, url credentials, and global `docker_isolation.extra_args` |
| `internal/httpapi/import.go` `runImport` | imported url/command/args raw (its CLI twin was already redacted) |
| `internal/server/mcp.go` `createDetailedErrorResponse` | the full upstream request URL, query credentials and all, plus the raw response body |
| `internal/serveredition/api/admin_handlers.go` | raw `[]*config.ServerConfig` written to JSON |
| registry source URLs (5 doors) | operator-configured endpoints that can carry `?token=` |
| `handleGetDockerStatus` `last_error` | a docker failure quotes the run command line, `-e API_KEY=…` and all |
| `GetStats`, diagnostics/doctor doors | name-rule-only, so a credential under a benign key survived |

`handleApplyConfig` / `handlePatchConfig` gained the matching bind-or-refuse, which also fixed a **pre-existing** corruption: `PATCH /api/v1/config` was persisting an echoed env mask over the live credential. Config reads now fail closed — 500 rather than serve unredacted if the walk cannot round-trip.

## Why this needed nine rounds

Every round after the fourth found the same shape: a rule applied at one door and not its sibling. Rounds 8 and 9 stopped fixing doors and built the inventory instead — a scan that derives config-bearing types at a fixpoint and fails when a new unrouted door appears.

That scan immediately proved its worth on the guards themselves:

- `TestRoutedDoors_StillCallASharedRule` was **vacuous**. Deleting `redactedServerViews` from `handleListQuarantinedUpstreams` — the original #1148 door — left it green, because an unrelated mention of `liveRedaction` elsewhere in the body satisfied it. Each row now names the rules it must route through, demonstrated red.
- Seven correctly-redacted doors were missing from the pinned inventory, so their redaction could have been deleted later with nothing failing.
- The renderer-registration gate only discovered package-level functions, while the renderings that reach live doors are **methods** on `Redaction`.

## Verification

Build, `-tags server`, `go test ./internal/... -race`, `-tags server ./internal/serveredition/...`, golangci-lint v2 → 0 issues.

Every fix was proven red on the unfixed tree first, and each guard was demonstrated going red when the thing it guards regresses.

Final review verdict: `readyToMerge: true`, four remaining findings all minor (a scan blind to three rarer raw-config shapes; a missing separator in one refusal message; a raw-value fallback in the import view when `json.Marshal` fails; and one pre-existing design note the reviewer explicitly flagged as not introduced here).

## Not fixed here

The server-edition per-user handlers are #1161, handled separately — **and the scan found a fifth handler that issue does not name: `enableServer`**. It lists only `listServers`/`getServer`/`createServer`/`updateServer`. All five are recorded in an `openDoors` table that requires an issue number per row and goes red once the fix lands, so the rows cannot quietly become permanent exemptions.


Dumbris:
## CodeQL red: pre-existing alert, not introduced here

The CodeQL check reports `1 new alert including 1 critical severity security vulnerability`. It is **not new**, and it is not in this PR's diff. Evidence:

```
alert 125  go/request-forgery  internal/telemetry/optout.go:135
created: 2026-06-15        (~3 months before this PR)
open on: refs/heads/main
         refs/pull/1164/merge
         refs/pull/684/merge
         refs/pull/784/merge
```

`internal/telemetry/optout.go` is not among the 24 files this PR changes; it was last touched by #1037 and #684. The alert is already open on `main`, so merging this does not change the repo's exposure — CodeQL's "in code changed by this pull request" attribution is what is wrong, most likely because this PR's dataflow-adjacent changes intersect the reported path.

The sink is also already mitigated at that line — an explicit CWE-918 barrier on the same control-flow path:

```go
if u.Scheme != "http" && u.Scheme != "https" { return "", ... }
host := strings.ToLower(u.Hostname())
switch host {
case "localhost", "127.0.0.1", "::1", defaultTelemetryHost():
    return u.String(), nil
default:
    return "", fmt.Errorf("telemetry URL host %q not allowed", host)
}
```

Scheme is constrained and the host must **exactly** equal an allowlisted value; the request is only issued on that branch. CodeQL is not recognising the switch as a sanitiser.

Flagging rather than quietly merging past it, since this is a security PR and a red critical check is exactly the thing that should not be waved through on a maintainer's behalf. If you would rather it be green first, the alert wants dismissing as a false positive on `main` (or the barrier reshaping into a form CodeQL recognises) — either way that is a change to `internal/telemetry`, which is out of scope here.

Issue #1173, opened 2026-09-02
[Bug]: internal_tool_call activity records skip the 64KB response truncation


antongulin:
## What happened?

My `config.db` grew to about 940MB. Most of it is the activity log.

When I looked inside, I found records far bigger than the 64KB response limit. The biggest single record is **10.75MB**, which is about 160 times the cap.

I checked every activity record over 200KB. There are 82 of them, and they are all the same kind:

```
types of oversized records: {'internal_tool_call': 82}

 SIZE MB  TYPE                   TOOL              response field len   response_truncated
   10.75  internal_tool_call     call_tool_read              10470961   False
    5.14  internal_tool_call     call_tool_read               5133653   False
    4.98  internal_tool_call     call_tool_read               4972547   False
    4.82  internal_tool_call     call_tool_read               4813426   False
    4.70  internal_tool_call     call_tool_read               4698651   False
```

Every one is `type: internal_tool_call` with `tool_name: call_tool_read`. Not a single normal tool call is oversized. So regular calls are being truncated correctly, and this one path is not.

Also note the last column. None of these records has `response_truncated` set, which fits: nothing truncated them, so nothing flagged them.

The truncation code itself looks fine. `internal/storage/activity.go` has `DefaultMaxResponseSize = 64 * 1024` and a `truncateResponse(response string, maxSize int)` that appends `...[truncated]`. It just does not seem to run on this path.

There is a matching comment in `internal/httpapi/activity.go`:

> Arguments, Response and Metadata are unbounded in practice (only Response is truncated, at 64KB)

And `internal/server/mcp.go` has `emitActivityInternalToolCallTruncated`, described as being for a handler whose recorded response is larger than what the agent received. So a truncating variant exists for the internal path. It looks like the plain internal emit path is the one that stores the full response.

I have not traced which call site is at fault, so I may have the mechanism wrong. The stored data is what I am sure about.

## Expected behavior

An `internal_tool_call` record should be truncated at `activity_max_response_size` (64KB by default), the same as a normal tool call, and should get `response_truncated: true` when that happens.

## Steps to reproduce

1. Add an upstream with a tool that returns a very large response. I hit this with `cloudflare-workers` `workers_get_worker_code`, which returns a whole bundled Worker.
2. Call that tool through mcpproxy's own `call_tool_read`, so the record is stored as `internal_tool_call`.
3. Read the newest record back from the `activity_records` bucket in `config.db`.
[...]

PR #1174, opened 2026-09-02
fix(activity): apply activity_max_response_size on the write path

antongulin:
# Pull Request

Related #1173

## Description

`activity_max_response_size` is declared in config with a 64KB default and **never read**. `truncateResponse` and its exported wrapper `TruncateActivityResponse` have **no production caller** either — the only caller in the tree is `internal/storage/activity_test.go:317`.

So nothing bounds a single activity record. Retention prunes the log by age, count and total size, but the full response text is persisted for every record type, and one proxied upstream response can be many megabytes.

On the deployment where I found this, `activity_records` held 82 records over 200KB, the largest **10.75MB** — about 160x the configured cap — with nothing logged to suggest anything was wrong. `config.db` had reached ~940MB.

### What this changes

It wires up the truncation that already exists, rather than adding a new mechanism:

- `maxResponseSize` on `ActivityService`, defaulting to the documented 64KB
- `SetMaxResponseSize`, taking the same "ignore non-positive" shape as the other retention setters, so an unset config value cannot silently disable the cap
- applied in the three places that build a record: `handleToolCallCompleted`, `handleInternalToolCall`, `handlePromptGet`
- fed `cfg.ActivityMaxResponseSize` from `runtime.New`, next to the existing retention wiring, and added to the "Activity retention config applied" log line

### Two things worth a reviewer's attention

**1. `ResponseTruncated` is OR-ed, not assigned.** That field already carries the Spec 103 meaning "the recorded response is *larger* than the one the agent received" (`activity_service.go`, the comment above `responseTruncated := getBoolPayload(...)`). Storage truncation is the opposite direction. I OR-ed the two so neither can clear the other, on the basis that the field documents itself as "true if response was truncated" and a consumer wanting the untruncated text is out of luck in both cases. If you would rather have a separate field for the storage-side meaning, say so and I will split it.

**2. `handlePromptGet` did not set `ResponseTruncated` at all** and now does.

There is also a small alias, `DefaultActivityMaxResponseSize = storage.DefaultMaxResponseSize`, because `NewActivityService`'s first parameter is named `storage` and shadows the package inside that function body.

## Testing

- [x] All existing tests pass

New file `internal/runtime/activity_response_truncation_test.go`, five cases. They drive the real event path through `handleEvent` and assert on what actually reached storage, rather than on a mock:

- an upstream tool-call response is capped and flagged
- an `internal_tool_call` response takes the same cap (this is `call_tool_read`, where the large records came from)
- an emitter-set `response_truncated` survives a record storage does *not* shorten
- the 64KB default applies with no config set
- `SetMaxResponseSize` ignores non-positive values

**Negative control:** with the body of `truncateForStorage` replaced by `return response, false`, three of the five fail; restoring it makes them pass. So they fail on pre-fix code, not just on a mock.

```
--- FAIL: TestToolCallResponseTruncatedForStorage
--- FAIL: TestInternalToolCallResponseTruncatedForStorage
--- FAIL: TestDefaultResponseCapAppliesWithoutConfig
```

Ran green:

```
ok  internal/runtime               122.985s
ok  internal/runtime/configsvc       0.139s
ok  internal/runtime/stateview       0.151s
ok  internal/runtime/supervisor      1.875s
ok  internal/runtime/supervisor/actor 0.839s
ok  internal/storage                24.752s
ok  internal/config                  6.242s
ok  internal/httpapi                 1.025s
```

Plus `-race` on the new tests, `go build ./...`, `go vet ./internal/runtime/...`, `gofmt`, and `staticcheck ./internal/runtime/` — all clean. `internal/server` was still running locally when I opened this; I will follow up in a comment with the result, and CI covers it regardless.

### A note for anyone writing similar tests

`storage.DefaultActivityFilter()` sets `ExcludeCallToolSuccess: true`, so a successful `call_tool_*` record is invisible to it. My first draft used that filter, saw zero rows and looked like a storage bug. The tests pass an explicit `storage.ActivityFilter{Limit: 100}` instead.

### DIFF
--- a/docs/features/activity-log.md
+++ b/docs/features/activity-log.md
@@ -474,7 +474,7 @@ Activity logging is enabled by default. Configure via `mcp_config.json`:
 | `activity_retention_days` | 90 | Days to retain activity records |
 | `activity_max_records` | 100000 | Maximum records before pruning oldest |
 | `activity_max_size_mb` | 256 | Maximum total activity-log size in MB before pruning oldest (`0` disables). Runs alongside the age and count caps to bound `config.db` growth when records carry large payloads. |
-| `activity_max_response_size` | 65536 | Max response size stored (bytes) |
+| `activity_max_response_size` | 65536 | Max response text stored per record (bytes). Applies to `tool_call`, `internal_tool_call` and `prompt_get`; `0` or absent falls back to 65536 and it cannot be disabled. Read at startup only, like its retention siblings. Truncated text keeps a `...[truncated]` suffix. Already-stored records are not rewritten, and (as with pruning) the BBolt file does not shrink on disk. Sensitive-data detection still scans the untruncated response, under its own `sensitive_data_detection.max_payload_size_kb` cap. |
 | `activity_cleanup_interval_min` | 60 | Background cleanup interval (minutes) |
 
 > **Why the size cap?** The age and count caps alone do not bound disk: with large per-record payloads the log can reach hundreds of MB while still under 100k records / 90 days. `activity_max_size_mb` removes the oldest records (always keeping the newest) until the log is within the byte budget. Note: pruning frees pages for reuse but does not shrink the database file on disk (BBolt does not return freed pages to the OS).
--- a/internal/runtime/activity_response_truncation_test.go
+++ b/internal/runtime/activity_response_truncation_test.go
@@ -1,0 +1,162 @@
+package runtime
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+
+	"github.com/smart-mcp-proxy/mcpproxy-go/internal/storage"
+)
+
+// allActivities lists without DefaultActivityFilter's ExcludeCallToolSuccess,
+// which would drop the successful call_tool_* records this file is about.
+func allActivities(t *testing.T, store *storage.Manager) []*storage.ActivityRecord {
+	t.Helper()
+	records, _, err := store.ListActivities(storage.ActivityFilter{Limit: 100})
+	require.NoError(t, err)
+	return records
+}
+
+// Retention bounds the activity log by age, count and TOTAL size, but nothing
+// bounded a SINGLE record: activity_max_response_size was declared in config and
+// never read, and TruncateActivityResponse had no production caller. A proxied
+// upstream response was persisted whole, so a handful of large calls could add
+// hundreds of MB to config.db with nothing logged.
+//
+// Each case drives the real event path and asserts against what actually landed
+// in storage, so it fails on the pre-fix code rather than on a mock.
+
+func TestToolCallResponseTruncatedForStorage(t *testing.T) {
+	store, cleanup := setupTestStorage(t)
+	defer cleanup()
+
+	svc := NewActivityService(store, zap.NewNop())
+	svc.SetMaxResponseSize(1024)
+
+	svc.handleEvent(Event{
+		Type:      EventTypeActivityToolCallCompleted,
+		Timestamp: time.Now().UTC(),
+		Payload: map[string]any{
+			"server_name": "upstream",
+			"tool_name":   "big_tool",
+			"status":      "success",
+			"response":    strings.Repeat("x", 50_000),
+		},
+	})
+
+	records := allActivities(t, store)
+	require.Len(t, records, 1)
+
+	assert.Less(t, len(records[0].Response), 50_000,
+		"a 50KB response must not be persisted whole under a 1KB cap")
+	assert.True(t, strings.HasSuffix(records[0].Response, "...[truncated]"),
+		"truncation must be visible in the stored text")
+	assert.True(t, records[0].ResponseTruncated,
+		"a record shortened on the write path must say so")
+}
+
+// call_tool_read proxies whole upstream payloads, which is where the largest
+// records in a real database came from.
+func TestInternalToolCallResponseTruncatedForStorage(t *testing.T) {
+	store, cleanup := setupTestStorage(t)
+	defer cleanup()
+
+	svc := NewActivityService(store, zap.NewNop())
+	svc.SetMaxResponseSize(1024)
+
+	svc.handleEvent(Event{
+		Type:      EventTypeActivityInternalToolCall,
+		Timestamp: time.Now().UTC(),
+		Payload: map[string]any{
+			"internal_tool_name": "call_tool_read",
+			"status":             "success",
+			"response":           strings.Repeat("y", 50_000),
+		},
+	})
+
+	records := allActivities(t, store)
+	require.Len(t, records, 1)
+
+	assert.Less(t, len(records[0].Response), 50_000,
+		"internal tool calls take the same cap as upstream calls")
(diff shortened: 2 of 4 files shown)


antongulin:
Following up on the `internal/server` run I said was still going when I opened this.

**It passes.** Zero failures, zero skips:

```
ok  github.com/smart-mcp-proxy/mcpproxy-go/internal/server         593.273s
ok  github.com/smart-mcp-proxy/mcpproxy-go/internal/server/tokens    1.156s
exit=0
```

Worth recording that my first attempt at this package was invalid, in case anyone reads the earlier state. I ran it without building the binary first, and 12 E2E tests failed instantly:

```
--- FAIL: TestBinaryStartupAndShutdown (0.00s)
    binary.go:154: mcpproxy binary not found at .../internal/server/mcpproxy.
    Set MCPPROXY_BINARY_PATH to the built binary or run: go build -o mcpproxy ./cmd/mcpproxy
```

That was me skipping the build step CONTRIBUTING calls for, not a regression from this change. After `go build -o mcpproxy ./cmd/mcpproxy` and re-running with `MCPPROXY_BINARY_PATH` set, the result above is what I get.

Full local verification for this branch:

```
ok  internal/runtime                  122.985s
ok  internal/runtime/configsvc          0.139s
ok  internal/runtime/stateview          0.151s
ok  internal/runtime/supervisor         1.875s
ok  internal/runtime/supervisor/actor   0.839s
ok  internal/storage                   24.752s
ok  internal/config                     6.242s
ok  internal/httpapi                    1.025s
ok  internal/server                   593.273s
ok  internal/server/tokens              1.156s
```

Plus `-race` on the new tests, `go build ./...`, `go vet ./internal/runtime/...`, `gofmt`, and `staticcheck ./internal/runtime/` — all clean.

So the third checkbox in the description is now backed by an actual full run rather than a partial one.

Dumbris (changes requested):
Thanks for this — the diagnosis is right, the wiring is the correct shape, and the negative control in the description is exactly what I want to see. Two things need to change before merge, and a few smaller ones.

### Blocking

**1. Sensitive-data detection now scans the truncated text.**
In `handleToolCallCompleted` the new line reassigns `response` before the Spec 026 block reads it (`scanText := response` at ~L706, when no `detection_text` is supplied — the default). `handlePromptGet` does the same with `responseStr` (~L1062). Before this PR the detector saw the full response and applied its own cap (`sensitive_data_detection.max_payload_size_kb`, default 1024KB); after it, anything past byte 65536 is never scanned, and nothing records that the scan window shrank. Please capture the untruncated string before truncating and pass that to `runAsyncDetection` in both handlers, and add a test with a secret placed after the 64KB boundary under the default cap.

**2. OR-ing storage truncation into `ResponseTruncated` breaks the cost predicate for `internal_tool_call` records.**
Two consumers read that flag on internal records with the specific meaning "the agent received *less* than `ResponseBytes`, so exclude the row": `internal/runtime/usage_aggregate.go:491` (`truncatedBuiltinOverstatesDelivery`, feeding the Usage timeline, rebuilt from persisted records on cold start) and `bench/replaycorpus/flags.go:373` (`ReasonTruncatedRetrieveOverstates`). A storage-cut, delivered-whole built-in now carries the flag with an honest `ResponseBytes`, so it is dropped from delivered traffic — reachable in default config for `describe_tool` (no agent-side cut) and for `retrieve_tools`/`code_execution` whenever `activity_max_response_size < tool_response_limit`. Reproduced: an 80,000-byte `describe_tool` record gives `RespBytesSum=0` on this branch and `80000` on `main`.

For `tool_call` and `prompt_get` the OR is fine — the comment above that predicate already defines the flag on upstream records as "cut on the way into the log, bytes honest". The narrow fix is to not OR for `ActivityTypeInternalToolCall` (the stored text still carries the `...[truncated]` suffix), plus a test mirroring `TestUsageAggregate_TruncatedBuiltinDoesNotInflateDeliveredBytes` in the other direction. A separate `storage_truncated` field is the cleaner answer but drags in contracts, swagger, CSV export and three UI badges; I would leave that for a follow-up.

### Non-blocking

- Please change `Fixes #1173` to `Related #1173` — this repo never auto-closes issues from PRs (see the commit conventions in `.specify/memory/constitution.md`), and in this case the reporter's table shows the other 432MB is `server_*_tool_calls`, which this PR does not touch (tracked as #1176). `Arguments` is also still stored whole on every record type; that can be a follow-on issue too.
- `handlePromptGet` truncation and its new flag have no test; a sixth case mirroring `TestToolCallResponseTruncatedForStorage` with `EventTypeActivityPromptGet` would close that (the whole package still passes with only the prompt path neutered).
- `docs/features/activity-log.md` row for `activity_max_response_size` could now say what it enforces: applies to `tool_call`/`internal_tool_call`/`prompt_get`, `0`/absent = 65536, cannot be disabled, applied at startup only (same as its retention siblings), already-stored records are not rewritten and the BBolt file does not shrink.

Everything else checked out: UTF-8 splitting at the cut is tolerated by every consumer (encoding/json coerces), the suffix can push a record a few bytes over the cap (fine), `-race` is clean, and I confirmed the negative control independently.

antongulin:
Both blockers fixed in 6901487.

**1. Scan before truncate.** `handleToolCallCompleted` and `handlePromptGet` now capture the response as received and pass that to `runAsyncDetection`, while still storing the cut copy. `detection_text` still wins where the TOON seam supplies it. `TestDetectionScansUntruncatedResponse` puts an AWS key past the 64KB boundary under the default cap and asserts it is still detected on both paths — and absent from the stored response.

**2. No OR for `internal_tool_call`.** `handleInternalToolCall` truncates the stored text without touching `ResponseTruncated`; `tool_call` and `prompt_get` keep the OR, where the meaning already matches. `TestStorageTruncatedBuiltinStillCountsDeliveredBytes` drives an 80,000-byte `describe_tool` through the event path and asserts 80,000 delivered bytes; `TestInternalToolCallResponseTruncatedForStorage` now asserts the flag stays false. Left `storage_truncated` as a follow-up, as you suggested.

Also done: `TestPromptGetResponseTruncatedForStorage` for the untested prompt path, PR body changed to `Related #1173`, and the `activity_max_response_size` doc row now states what it enforces, which record types it covers, startup-only, no rewrite of stored records, and that detection is unaffected.

Both fixes negative-controlled — restoring either defect fails the new tests and nothing else. `./internal/runtime/...` and `./internal/storage/...` pass.

One thing I left alone: the first commit's message still reads `Fixes #1173`. Say the word and I will amend and force-push.

Dumbris:
Thanks Anton — merged.

Good catch. A config field declared and never read, plus an exported helper with no production caller, is exactly the kind of thing that sits there for a year. The 10.75MB record and the 940MB `config.db` made the case better than any argument would have.

Both review fixes are right, and I checked the negative controls rather than taking them on trust — reverting either one does fail the new tests.

One thing: `TestDetectionScansUntruncatedResponse` times out under `-race`. Its ~75KB payload takes ~9.5s to scan (the race detector costs ~24x on that path) against `waitForDetectionMetadata`'s 5s deadline — which is why it went red in CI and green for you locally. Nothing wrong with the logic; just the payload sizing. Fixed in #1207 with an explicit 1KB cap instead of the 64KB default: same seam, 0.5s.

Thanks also for the `DefaultActivityFilter` note in the description — that one would have cost the next person an hour.

PR #1199, opened 2026-09-02
chore: remove accidentally committed warmtiktoken binary, guard against repeats

Dumbris:
## What

`warmtiktoken` (43 MB arm64 Mach-O — the output of `go build ./bench/cmd/warmtiktoken` run from the repo root) was committed at the repo root in #1153. This PR:

- untracks `/warmtiktoken` and ignores it (plus `/mcpproxy-server`) alongside the other root binaries in `.gitignore`
- adds `check-added-large-files` (8 MB cap; largest legitimate screenshot/report is ~6.5 MB) to pre-commit
- adds a local `no-native-binaries` pre-commit hook (`scripts/check-no-native-binaries.sh`) that rejects Mach-O/ELF/PE files outright

Both new hooks ran on this commit; the script self-test rejects `warmtiktoken` and passes text files.

## Repo audit

- `warmtiktoken` was the only compiled executable tracked, and the only tracked file > 8 MB.
- No `.env`, key, cert, DB, log or config-with-credentials files are tracked.
- Secret patterns (`mcp_agt_`, `sk-*`, `ghp_`/`github_pat_`, `AKIA`, `xox*`, private-key headers, `AIza`, `X-API-Key`/`apikey=` with real-looking values) grep clean across the tree: every hit is a documented example, a detector fixture, or a test-only key such as `e2e-test-api-key-12345`.
- A full-history `gitleaks git` scan result is noted in a follow-up comment.

## Not done here

The 43 MB blob remains in history. Removing it needs `git filter-repo` + a force-push of `main` and every open branch, so that is a separate decision.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@
 *.so
 *.dylib
 /mcpproxy
+/mcpproxy-server
+/warmtiktoken
 /test/launcher-server/launcher-server
 __debug_bin*
 
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ __debug_bin*
 # Playwright MCP artifacts
 .playwright-mcp/
 
+# osgrep local server config (contains a local auth token)
+.osgrep/
+
 # Bleve search index directories
 index.bleve/
 *.bleve/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,9 +15,21 @@ repos:
         exclude: '\.(patch|diff)$'
       - id: end-of-file-fixer
       - id: check-merge-conflict
+      - id: check-added-large-files
+        # A 43 MB `warmtiktoken` Mach-O binary was committed at the repo root
+        # in #1153 (go build in the wrong cwd). Screenshots and verification
+        # reports are legitimately 1-6 MB, so cap well above them.
+        args: ['--maxkb=8192']
 
   - repo: local
     hooks:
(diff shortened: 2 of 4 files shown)


Dumbris:
## Full-history secret scan (gitleaks 8.x, all refs)

520 raw findings, all triaged:

| Bucket | Count | Verdict |
|---|---|---|
| `internal/security/patterns/*`, detector/masking tests, docs, spec quickstarts, e2e stubs | ~505 | Intentional fixtures and documented examples (`AKIAIOSFODNN7EXAMPLE`, `mcp_agt_a1b2c3d4…`, PEM headers, `*-12345` keys). |
| `test/e2e-config.json` `api_key` | 3 values over time | Auto-generated 64-hex key for a localhost test instance, committed since #477. Not Algis's live key. Replaced with `e2e-test-api-key-fixture` in e8b1d1c (the e2e script extracts the key from the core's log, so the value never mattered). |
| `.osgrep/server.json` `authToken` | 1 | Local osgrep server token, committed only on the unmerged `pull/165` ref (never on `main`). `.osgrep/` is now ignored. |
| CI/scripts `X-API-Key` headers | ~10 | `qa-sandbox-ci-test`, `wrong-api-key`, `test-api-key` — synthetic. |

No credential for an external service (GitHub, Anthropic, cloud, Slack, Stripe, Cloudflare) appears anywhere in history.

Issue #1076, opened 2026-08-28
Web UI: quarantined server shows a red "file a bug report" error panel above its own calm quarantine banner


Dumbris:
## Summary

On Server Detail, a quarantined server renders a red `ErrorPanel` **and** the calm Security Quarantine banner at the same time, off one payload that says the server is `healthy` / `quarantined`. The error panel is usually the unclassified one, so the user is told to file a bug report about a server that is simply waiting for them to approve it.

This is the same defect [#1075](https://github.com/smart-mcp-proxy/mcpproxy-go/pull/1075) fixed for the macOS tray badge, on a surface that PR deliberately did not touch.

## Where

[`frontend/src/views/ServerDetail.vue:1809`](https://github.com/smart-mcp-proxy/mcpproxy-go/blob/main/frontend/src/views/ServerDetail.vue#L1809)

```js
const showDiagnosticPanel = computed(() => {
  const d = server.value?.diagnostic
  if (!d || !d.code) return false
  return d.severity === 'warn' || d.severity === 'error'
})
```

No quarantine exemption. And the quarantine banner at `:217` is a **separate** `v-if`, not part of the `v-if` / `v-else-if` chain at `:181`–`:201`, so both blocks render together.

## Why the payload looks contradictory

mcpproxy still attempts a connection to a quarantined server so the security scanner can export its tool definitions. That attempt fails, leaving an error-severity diagnostic behind — while `internal/health/calculator.go` short-circuits quarantined servers to `level: healthy`, `admin_state: quarantined`, `action: approve`. Both facts are true in one response; the UI reads only the first.

Real payload from a live install:

```json
{
  "name": "everything",
  "quarantined": true,
  "diagnostic": {
    "code": "MCPX_UNKNOWN_UNCLASSIFIED",
    "severity": "error",
    "user_message": "mcpproxy could not classify this failure. Please file a bug report so we can add a specific code.",
    "fix_steps": [{ "type": "link", "label": "Report a bug", "url": ".../issues/new" }]
  },
  "health": {
    "level": "healthy",
    "admin_state": "quarantined",
    "summary": "Quarantined for review",
    "action": "approve"
  }
}
```

## Reproduce

1. Add an MCP server that cannot connect (an unreachable URL is enough) and leave it quarantined.
2. Wait for the admission scan to attempt its inspection connect and fail.
3. Open Server Detail for it.

[...]

PR #1216, opened 2026-09-06
fix(webui): stop the UI asserting states it was never told (v0.65.0 UX audit)

Dumbris:
## What this is

Fixes for the findings of a first-time-developer UX audit of MCPProxy v0.65.0. One theme runs through nearly all of them: **the Web UI stated things it had not been told, or described a state with words that meant the opposite of the truth.** A green "Healthy" on a server no client can reach. A "Round-trip verified" finish line stamped off a transport handshake. A "Setup ✓" that was really "we never fetched". "No tools available — server must be connected" on a server whose tools were deliberately withheld for security review.

Almost all of it is copy and render-gating in `frontend/src/`. There is exactly one backend change (7 lines in `internal/server/server.go`) and one Swift change that follows from it.

This is an integration of six focused branches, each pushed separately if anyone would rather review them one at a time: `fix/webui-readiness-and-review-surface`, `fix/webui-search-expectations`, `fix/webui-add-server-handoff`, `fix/webui-setup-verify-honesty`, `fix/webui-auth-and-loading-state`, `fix/duplicate-endpoint-clarity`.

---

### F08/F10/F12 — the server detail screen said "absent" where the truth was "withheld", "off", or "nothing written"

**Health tile (F10).** `internal/health/calculator.go` deliberately reports a quarantined or disabled server as level `healthy` — "quarantined is intentional, not broken". The tile printed that *level*, so it rendered a green **Healthy** on precisely the servers no client can reach, and could pair it with a "Sign-in required" sub-line, a combination only reachable on a quarantined/disabled server. The tile's single word now leads with the admin state (**Blocked** / **Off**); the observed health keeps its own vocabulary on the sub-line beneath, which already printed `health.summary`. Neither reads `text-success` any more. The hard-coded "set by you" caption under Admin state was false for the *default* quarantine-on-add path, so a quarantined server now reads "awaiting your review".

**Tools tab empty state (F08).** A server quarantined at admission is `connected:false` and its tools are kept out of both the state snapshot and the search index by design (`internal/runtime/tool_quarantine.go`). `GET /servers/{id}/tools` therefore returns `[]`, and the tab rendered the same generic "No tools available / Server must be connected" a genuinely broken server gets — while the Security tab was asking the operator to review those very definitions. It is now quarantine-aware ("Tools withheld for review") with a button through to the Security tab. The sentence deliberately names **no count**: `quarantine.pending_count` counts tools *awaiting review*, which is not the count of tools *withheld* (quarantining a trusted server with 20 approved and 2 new tools reports 2 while all 22 are withheld), and no payload field carries the real total.

**Missing per-server log file (F12) — the one backend change.** `GET /api/v1/servers/{id}/logs` returned HTTP 500 `"log file not found … (server may not have run yet)"` for any server with no log file. That claim is false: the file is created lazily by lumberjack and is level-gated (`internal/logs/logger.go`), so a connected, healthy server that logged nothing at or above the configured level has no file at all — on any transport. The ENOENT branch now returns an empty (non-nil, so it marshals as `[]`) result and the UI falls into its normal empty state. Every genuinely broken case — unknown server, path escape, unreadable file — still errors.

### F11 — tool search blamed the query for what the filters had done

Typing "React documentation" returned nothing while "documentation" returned Context7: the filter matched the **whole query as one contiguous substring of a single field**. Terms are now matched independently (AND across terms, OR across fields) — a strict superset of the old behaviour, so no result that matched before can disappear. Still a substring filter, not BM25; wiring the box to `/api/v1/index/search` is a product decision and is reported, not built.

The empty state now names the query and **the scope it actually searched** — how many tools, across how many servers, whether disabled tools were in scope — and, only when there are any, how many quarantined servers are not listed here. Quarantined servers contribute zero tools to `GET /api/v1/tools` and quarantine is on by default, so a first-run developer was searching a silently truncated catalogue with no hint of it.

Reporting `stats.total` would have lied whenever another filter was also narrowing the list, so the filter chain is split into `searchScope` (every filter except the search box) and `filteredTools`. All filters are ANDed, so the result is identical and the numbers are now literally true. A separate branch handles the case where a status filter left **nothing** to search — no wording of the query can help there, so the empty state names the filter as the cause instead.

### F07 — Add Server showed dead controls, then abandoned the user

The Docker isolation toggle rendered disabled for URL upstreams, where isolation is structurally impossible (`internal/config/isolation_resolve.go` resolves `IsolationSourceNotStdio` whenever Command is empty). It is now `v-if="type === 'stdio'"`. **Presentation only** — both existing state guards are untouched, so `isolation_json` still cannot reach an http add. The permanently-disabled "Coming soon" *Idle on Inactivity* toggle is removed; `handleSubmit` never read it and no backend field of that name exists.

After a successful add the user was dropped back where they started while the activation work (connect, scan, review, approve) waited on a page they had to find. The `added` event now carries the server name and the three navigating consumers push to `serverDetailPath(name)`. The navigation lives in the **consumers, not in `handleSubmit`** — `OnboardingWizard.vue` also listens to `added`, and a push from inside the modal would unmount the wizard mid-flow. The bulk-import path keeps its bare `emit('added')`: a bulk import has no single destination.

### F13 — the setup wizard called a handshake a verified round-trip

The Verify tab showed a green **"Round-trip verified"** off a single boolean, `first_mcp_client_ever`, which is stamped from the MCP `initialize` hook — a transport handshake, before any tool is listed let alone called. All three suggested prompts targeted mcpproxy's own built-ins, which record as `internal_tool_call` and never stamp `first_real_tool_call_ever`. A user who had never seen an upstream tool run was told the product was verified.

It is now two separately-labelled milestones: "AI client connected" (same condition, honest words) and "First upstream tool call". One suggested prompt was added that actually dispatches to an upstream. The copy says "ran through mcpproxy and returned a result", not "succeeded", because `recordRealToolCallSuccess()` is called one statement *before* the `isError` classification.

**And then the fix was corrected for the same defect in the opposite direction.** `first_real_tool_call_ever` has exactly one call site (the `call_tool_*` handler), so under `routing_mode=direct` or `code_execution` the new row told a user who had made dozens of real upstream calls "No upstream tool call yet", forever. Shipping a green claim off a handshake and a red claim off a blind flag are the same defect wearing different colours. The state is now tri-state and routing-mode aware, read from the same `GET /api/v1/status` response that already carries `activation` — **satisfied** (a lifetime fact, true under any mode), **pending** (only under the mode that actually stamps it), **unknown** (the row and its caveat do not render at all). The backend gap is real and still open; it needs Go tests and its own diff.

### F06 — the first load asserted state it had not been told

Behind the "Authentication Required" modal every surface still rendered, and three of them stated as fact things the UI had never been able to read. On one fresh install, unauthenticated vs authenticated at the same moment:

```
before, unauthenticated  Setup ✓ | MCPPROXY stopped | 0 connected | 0 tools
authenticated, same box  Setup 2 | MCPPROXY active  | 0 connected | 0 tools
```

The "Setup ✓" was the worst: `incompleteTabCount` reads `state?.incomplete_tab_count ?? 0`, so "never fetched" and "nothing left to do" both arrive as 0 — and since nothing moves `state` off null, this was **permanent** for any install whose fetch kept failing, not a pre-fetch flash. Three gates were added using flags that already existed: `setupStateKnown`, `serversStore.loaded`, and `systemStore.status !== null`. Unknown renders a plain label or an em-dash rather than a confident zero.

The auth screen's recovery instructions were also corrected, each verified against the code and by running it: `mcpproxy doctor` never prints the API key (`doctor_cmd.go` has zero matches for `apikey`/`api_key`) — replaced with `mcpproxy status`; the startup log carries the key in full only on the run that generated it; the tray labels are named without the OS claim they previously carried (the tray is a **build** split, `//go:build !nogui && !headless && !linux`, not an OS split — a macOS tarball or Homebrew user sees "Open Web Control Panel", only the DMG's Swift bundle says "Open Web UI in Browser").

### F09 — a duplicate endpoint was accepted in silence, then the gate contradicted itself

Adding a second upstream at an already-configured endpoint is accepted silently, and the first the user hears of it is the scanner reporting the **new** server as `dangerous`: two entries on one endpoint expose byte-identical tool names and descriptions, so `detect.shadowing.cross_server` emits a hard-tier "possible impersonation" finding per tool. Reproduced end to end on a scratch instance: a single `POST /api/v1/servers` of the duplicate produced `status: dangerous, finding_counts {dangerous: 2}` unprompted, and `POST …/security/approve` without force returned `409 server has 2 dangerous (hard-tier) finding(s)`.

Two halves, both frontend-only:

1. **Add-time observation.** The manual tab now names the configured server already holding the endpoint being typed, as a neutral note. Submit stays enabled; nothing is merged, reused or blocked. Matching is exact string equality after trim — `url` for http; command, the full args list **and `working_dir`** for stdio (`node server.js` in two directories is two different programs), with a negative transport guard so a stale `url` left on a converted stdio entry cannot be named as the occupant of a typed HTTPS endpoint. It can miss: the server list masks credential-shaped values, so a secret-bearing endpoint never matches. A hint, not a guarantee. **Web-UI door only** — the MCP `upstream_servers add` handler, the CLI and the import path still say nothing; covering them needs a warning field on the add response.

2. **Gate wording.** One render carried both "Informational — these findings do not block the tool" and "Force-approving bypasses the scanner gate", about the findings behind that 409. The first sentence was false — its source read `scan_informational.go`'s "drives NO gating whatsoever" as a claim about the *findings*, when it is a claim about the *trigger*. The panel receives neither `quarantined` nor the effective trust mode, so it now claims nothing about this server's state, only what is unconditionally true of the tier. The dialogs stop saying "the scanner gate", a phrase the UI never defined. And `ScanReport`'s Force Approve confirmation counted the wrong number — raw `summary.critical` while its own gate predicate reads `finding_counts.dangerous` — so it offered to bypass the gate "despite 0 critical finding(s)" for the two findings the backend was refusing on.

**Not changed:** `detect/checks/shadowing.go`. Making the clone branch endpoint- or identity-aware is a maintainer decision. Nothing here touches `isBlockingFinding`, `ApproveServer`, the quarantine default, or the force path — security posture is unchanged. Also not fixed: after force-approving a duplicate, the **original** server flips to `dangerous` on its next scan and the two permanently accuse each other. The add-time note only reduces how often a user gets there.

### Two cross-branch defects, visible only once the branches were merged together

1. A live probe found a quarantined stdio server with a nonexistent command reporting `status=error, connected=false`, yet `health.level` came back `healthy` with action `approve` (the calculator returns early on Quarantined), and every fault alert on the page is deliberately suppressed while quarantined (#1076). Net effect: the Tools tab said "approve the server to list them" and nothing contradicted it — approving would hand the user a second failure. The copy now names the connection blocker **only when the server is actually disconnected**; a connected quarantined server keeps the original sentence, pinned by a control test.

2. The Go change above makes a missing log file an empty *success*. The macOS tray's `loadLogs()` read its on-disk fallback only from `catch`, so that empty success silently skipped the fallback and dropped historical local logs from the view. It now falls back on an empty result too.

---

## Verification

- `go build ./...`, `go vet ./internal/...` — clean
- `golangci-lint run --config .github/.golangci.yml ./internal/...` (v2, the config CI uses) — **0 issues**
- `go test ./internal/server/ -skip "E2E|Binary|MCPProtocol"` — ok, 196s
- `cd frontend && ./node_modules/.bin/vitest run` — **1,247 tests / 119 files pass**
- `npm run type-check` (vue-tsc --noEmit) — clean
- Driven live on a built instance, including the exact audit shapes: a quarantined http upstream (page reads Blocked / Quarantined for review / Tools withheld for review), a connected fixture server with no log file (`200 {"logs":[],"count":0}` where it previously returned the 500), and the duplicate-endpoint add against a backend returning exactly the count the confirm now quotes.

Every behavioural change is backed by a test that was **run and observed to fail first**, and the new specs each carry assertions that passed before the change so they are not vacuous. Several were additionally bite-checked by neutering the fix in the working tree and confirming a red suite — that check is what caught two hunks (the `toolsEmptyHeading` change and the approve-dialog paragraph) that had no assertion behind them at all and were free to drift back.

No API payload field changed, so no `contracts.ts` regeneration and no `make swagger` (`make build` re-run: zero diff). No new dependencies.


Dumbris:
### The Windows build failure is pre-existing, not from this PR

`Build Binaries (windows-latest, windows, amd64, …)` failed in the **Run tests** step (the step summary points at "Build binary and create archives", which is misleading — the log ends `FAIL` → `exit code 1` inside the test step). The four sibling matrix jobs that also show red were **cancelled** by fail-fast, not independently failing.

The single failing test:

```
--- FAIL: TestCompactLockReleaseDoesNotRemoveSomeoneElsesLock
FAIL	github.com/smart-mcp-proxy/mcpproxy-go/cmd/mcpproxy	14.968s
```

**Not attributable to this branch:** this PR changes no file under `cmd/mcpproxy` (`git diff --name-only main...HEAD` has zero matches there). The test arrived on 2026-09-05 in f95fc1428 (#1214), and neither of the last two green `main` runs contained the file, so it has not yet run on Windows on `main`.

**Root cause.** `acquireCompactLock` stamps the lock with:

```go
```

The test acquires A and B from the same process, microseconds apart. Windows' wall clock has ~15.6 ms granularity, so `time.Now()` returns the **same** value for both and `token_A == token_B`. `releaseA()` then finds a matching token and deletes B's lock, which is exactly what the test asserts must not happen. Linux and macOS have nanosecond-resolution clocks, so their tokens differ and the test passes — hence Windows-only.

A dedicated task is fixing this on `main`; it is deliberately not being fixed here to keep this PR scoped to the UX audit findings.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #1217, opened 2026-09-06
fix(db): make the compaction lock token unique independent of the clock

Dumbris:
## Problem

`TestCompactLockReleaseDoesNotRemoveSomeoneElsesLock` fails on the `windows-latest` **Build Binaries** job of **PR Build**, blocking every PR. It is pre-existing and unrelated to any branch: introduced by f95fc1428 (#1214), and [#1215](https://github.com/smart-mcp-proxy/mcpproxy-go/pull/1215) — which touches only `.github/RELEASE_NOTICE.md` — failed with the byte-identical message (run 33946755580, job 101259930963).

## Root cause

`acquireCompactLock` stamped the lock file with

```go
```

and release compares `string(current) != token` to decide whether the lock is still ours.

On Linux/macOS `time.Now()` has nanosecond resolution, so two acquisitions in one process always differ. Windows' system clock is coarse (~0.5–15 ms); the test acquires A, removes the file, and acquires B back-to-back with no work between them, so both land in the same tick. Same PID + same timestamp = identical token, A's release believes B's lock is its own, and deletes it. The test running in 0.00s matches.

This is a real bug, not just a flaky test: the token is the only thing standing between a stale-lock cleanup and two concurrent compactions on the same BBolt database, and on Windows it can genuinely collide.

## Fix

Add a 16-byte `crypto/rand` nonce to the token. `pid` and `started` stay — they are what a human reads out of a stale lock file. The lock protocol and the compaction flow are untouched.

## Test

`TestCompactLockTokenIsUniqueWithinOneClockTick` acquires, releases, re-acquires, and compares the two lock files **with the clock-derived `started=` line stripped** — so it reproduces what Windows sees and bites on a nanosecond clock too, instead of passing vacuously everywhere but Windows.

Verified failing before the fix:

```
db_cmd_test.go:578: two acquisitions produced the same clock-independent token "pid=45729";
on a coarse clock A's release would delete B's lock
```

and passing after, alongside the original test. `go test ./cmd/mcpproxy/` green, `-race` green, `GOOS=windows go test -c` builds, `golangci-lint run --config .github/.golangci.yml ./cmd/mcpproxy/...` reports 0 issues.

### DIFF
--- a/cmd/mcpproxy/db_cmd.go
+++ b/cmd/mcpproxy/db_cmd.go
@@ -292,6 +292,11 @@ func runDBStats(cmd *cobra.Command, _ []string) error {
 	return nil
 }
 
+// compactLockNow is a seam so tests can freeze the clock. A frozen clock is
+// exactly the Windows condition this token guards against, which is what lets
+// the lock tests reproduce it on every platform instead of by luck.
+var compactLockNow = time.Now
+
 // compactDatabase rewrites srcPath in place via a temp file + atomic rename.
 //
 // The rename is what makes this safe to interrupt: until it succeeds the
@@ -339,7 +344,7 @@ func acquireCompactLock(dir string) (release func(), err error) {
 		return nil, fmt.Errorf("failed to generate compaction lock token: %w", errors.Join(rerr, cerr))
 	}
 	token := fmt.Sprintf("pid=%d\nstarted=%s\nnonce=%x\n",
-		os.Getpid(), time.Now().UTC().Format(time.RFC3339Nano), nonce)
+		os.Getpid(), compactLockNow().UTC().Format(time.RFC3339Nano), nonce)
 	_, writeErr := f.WriteString(token)
 	closeErr := f.Close()
 	if writeErr != nil || closeErr != nil {
--- a/cmd/mcpproxy/db_cmd.go
+++ b/cmd/mcpproxy/db_cmd.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"crypto/rand"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


Dumbris:
Pushed 9feb4ecab, porting the test seam from #1221 (opened as a duplicate of this PR, now closed).

The fix itself is unchanged. What this adds is that `TestCompactLockReleaseDoesNotRemoveSomeoneElsesLock` — the test that exposed the bug — was still clock-dependent: on a nanosecond-resolution platform the two acquisitions get different timestamps, so it passed for the wrong reason and only failed on `windows-latest`, intermittently.

`compactLockNow` is now a seam, frozen in both lock tests. A frozen clock *is* the Windows condition (coarse granularity, two acquisitions back-to-back), so the collision is exercised on every platform rather than by luck. That also lets `TestCompactLockTokenIsUniqueWithinOneClockTick` compare whole tokens instead of stripping the `started=` line to guess what Windows would collide on.

Mutation-checked on macOS — with the nonce reverted, both tests fail, the pre-existing one with the exact windows-latest message:

```
--- FAIL: TestCompactLockReleaseDoesNotRemoveSomeoneElsesLock
    A's release removed B's lock, so a third compaction could start: stat …/config.db.compact.lock: no such file or directory
--- FAIL: TestCompactLockTokenIsUniqueWithinOneClockTick
    two acquisitions produced the same token "pid=47741\nstarted=2026-09-05T12:00:00Z\n"
```

Also verified: `go test -count=1 ./cmd/mcpproxy/` green, `-race` green, `GOOS=windows go build ./...` and `go test -c` build, golangci-lint v2 with `.github/.golangci.yml` → 0 issues.

PR #1221, opened 2026-09-06
fix(db): make the compaction lock token unique independent of the clock

Dumbris:
## Problem

`TestCompactLockReleaseDoesNotRemoveSomeoneElsesLock` fails intermittently on the `windows-latest` "Build Binaries" job. It is not branch-specific — it failed on [#1215](https://github.com/smart-mcp-proxy/mcpproxy-go/pull/1215) (a docs-only change) and on [#1216](https://github.com/smart-mcp-proxy/mcpproxy-go/pull/1216), then passed on the next run of the same branch.

It is also a real correctness bug, not just a flaky test.

`acquireCompactLock` stamps the lock file with:

```go
```

and release compares `string(current) != token` to decide whether the lock is still its own. On Windows the system clock granularity is coarse (~0.5-15ms), so two acquisitions in the same process with no work between them can read the *same* `time.Now()`. Same pid + same timestamp = identical token, so A's release believes B's lock is its own and deletes it — freeing the pathname for a third compaction to run alongside B on the same BBolt database. Whether the two calls straddle a clock tick is what made it intermittent; the observed failures run in 0.00s.

## Fix

Add 16 `crypto/rand` bytes, hex-encoded, to the token. `pid` and `started` stay for whoever has to diagnose a stale lock file. The lock protocol and the compaction flow are unchanged.

## Testing

`compactLockNow` is a seam so tests can freeze the clock — which *is* the Windows condition, made deterministic.

- The existing `TestCompactLockReleaseDoesNotRemoveSomeoneElsesLock` now freezes the clock, so it exercises the collision on every platform instead of by luck. With the nonce reverted it reproduces the byte-identical windows-latest failure on macOS: `A's release removed B's lock, so a third compaction could start: stat …config.db.compact.lock: no such file or directory`.
- New `TestCompactLockTokenIsUniqueWithoutClockAdvance` asserts the token property directly (acquire → read file → release → re-acquire → read file → compare).

Both fail without the nonce and pass with it. Verified: `go test -count=5 ./cmd/mcpproxy/` (green), `go test -race` (green), `GOOS=windows go build ./...` and `GOOS=windows go test -c` (both build), `golangci-lint run --config .github/.golangci.yml ./cmd/mcpproxy/...` → 0 issues.

Windows CI on this PR is the remaining confirmation.

### DIFF
--- a/cmd/mcpproxy/db_cmd.go
+++ b/cmd/mcpproxy/db_cmd.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"crypto/rand"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"os"
@@ -291,6 +293,10 @@ func runDBStats(cmd *cobra.Command, _ []string) error {
 	return nil
 }
 
+// compactLockNow is a seam so tests can freeze the clock and prove the token
+// stays unique when the timestamp cannot distinguish two acquisitions.
+var compactLockNow = time.Now
+
 // compactDatabase rewrites srcPath in place via a temp file + atomic rename.
 //
 // The rename is what makes this safe to interrupt: until it succeeds the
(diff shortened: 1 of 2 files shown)


Dumbris:
Duplicate of #1217, which has the same crypto/rand nonce fix and got there first. #1217 also handles the failure path correctly (it closes the file handle before `os.Remove`, which matters because Windows refuses to delete an open file).

The one thing this PR had that #1217 lacked — the `compactLockNow` seam that freezes the clock so the pre-existing `TestCompactLockReleaseDoesNotRemoveSomeoneElsesLock` reproduces the Windows collision on every platform — is ported over in 9feb4ecab. Closing in favour of #1217.

PR #1234, opened 2026-09-10
fix(windows): clean up child processes via Job Objects

LocoLoboZ:
## Problem

`internal/upstream/core/process_windows.go` and `internal/upstream/launcher/launcher_windows.go` were both TODO stubs. `killProcessGroup` did nothing, and `terminateProcess`/`killProcess` only ever killed the one PID mcpproxy itself spawned (e.g. `cmd.exe`), never the grandchildren it spawns (`node.exe`, `python.exe`, the actual MCP server process).

Every stdio server restart, reconnect, or disconnect on Windows leaks its grandchild process tree permanently. Confirmed **413 orphaned `node.exe`/`python.exe`/`qmcp.exe` processes** accumulated from under an hour of normal use with ~15 configured servers on one machine.

## Fix

Wraps each spawned process in a Windows Job Object configured with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` (new `internal/winjob` package). Job membership is inherited by anything the process spawns afterwards, so closing the job reaches the whole tree in one call — the Windows equivalent of `kill(-pgid, ...)` on the existing Unix path.

Two details that mattered in testing:

1. **The job must be assigned right after `cmd.Start()` succeeds, before the MCP `initialize()` handshake** — not after a successful handshake. Assigning only on success misses every leak caused by a restart/disconnect interrupting an in-progress connect attempt, which is the dominant real-world case (a slow `npx` download, bulk server-add contention, a timeout mid-handshake). This was the actual cause of most of the 413 leaked processes observed.
2. `golang.org/x/sys/windows` (as vendored here, v0.47.0) does not export `Set/QueryInformationJobObject`, only `CreateJobObject`/`AssignProcessToJobObject`/`TerminateJobObject`. `winjob.go` binds `SetInformationJobObject` directly via `NewLazySystemDLL` rather than adding a new dependency for one call.

Same fix applied to the separate `launcher` package (HTTP/SSE launcher-managed servers with `Command` set), which had the identical gap and an explicit `// left to a follow-up (matching the TODO already in internal/upstream/core/process_windows.go)` comment pointing at this file.

## Verification

- Standalone reproduction: spawn `cmd.exe` → `ping.exe` (grandchild), assign to job, `Close()` → grandchild confirmed killed.
- 6 sequential real restarts of a stdio server (`mcpproxy upstream restart <name>`, few seconds apart) → **0 leaked processes**, where every restart leaked one before this fix.
- `go build`/`go vet` clean for `windows/amd64`.
- Existing `internal/upstream/launcher` test suite passes unmodified.

## Known remaining gap (out of scope for this PR)

Firing many restarts of the *same* server at the exact same instant (an artificial stress case constructed for testing, not normal operation — mcpproxy's own reconcile loop restarts sequentially) can still leak a few processes. That traces to a separate race on the shared `processGroupID` field across concurrent `Connect`/`Disconnect` calls on one `Client`, is not Windows-specific, and is not addressed here.

### DIFF
--- a/internal/upstream/core/connection_stdio.go
+++ b/internal/upstream/core/connection_stdio.go
@@ -260,6 +260,21 @@ func (c *Client) connectStdio(ctx context.Context) error {
 		return fmt.Errorf("failed to start stdio client: %w", err)
 	}
 
+	// Extract the process group ID (Windows: assign to a Job Object) IMMEDIATELY
+	// after Start() succeeds, not after initialize() below. A restart/disconnect
+	// that interrupts an in-progress handshake (slow npx download, bulk-add
+	// contention, a timeout) previously left processGroupID at its zero value
+	// the whole time initialize() was running, so the init-failure cleanup path
+	// a few lines down (`if c.processGroupID > 0 { killProcessGroup(...) }`)
+	// silently did nothing and the spawned process (and on Windows, everything
+	// IT spawns) leaked. Setting it here means that cleanup path — and any
+	// concurrent restart that arrives before initialize() returns — has a real
+	// PID/Job Object to kill. The later extraction below is now just a no-op
+	// (its own `if c.processGroupID <= 0` guard) for the normal success path.
+	if c.processCmd != nil && c.processCmd.Process != nil {
+		c.processGroupID = extractProcessGroupID(c.processCmd, c.logger, c.config.Name)
+	}
+
 	// CRITICAL FIX: Enable stderr monitoring IMMEDIATELY after starting the process
 	// This ensures we capture startup errors (like missing API keys) even if
 	// initialization fails with timeout. Previously, stderr monitoring started
--- a/internal/upstream/core/process_windows.go
+++ b/internal/upstream/core/process_windows.go
@@ -4,9 +4,13 @@ package core
 
 import (
 	"context"
+	"os"
 	"os/exec"
+	"sync"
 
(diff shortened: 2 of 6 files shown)


Dumbris:
Thank you @LocoLoboZ — this is a real and long-standing gap (the Windows stubs were TODOs since the launcher landed), and the point about assigning the Job right after `Start()` rather than after the handshake is exactly right; it is what makes the init-failure and connect-failure cleanup paths reach the whole tree.

Review notes, so you know what happens next: the Job is closed only from `killProcessGroup`, which core's `Disconnect` reaches only when the graceful close fails — and mcp-go's `Stdio.Close()` returns within ~8 s (< the 10 s budget), so on the normal restart/disconnect path the Job is never closed and EOF-ignoring grandchildren still survive. Likewise the launcher's `reap()` waits on the log pumps before `job.Close()`, and a grandchild holding the inherited stdout pipe keeps the pumps from ever seeing EOF. Both are follow-ups, not blockers — this PR is a strict improvement (and `x/sys/windows` already exports `SetInformationJobObject` / `JOBOBJECT_EXTENDED_LIMIT_INFORMATION`, so the hand-rolled bindings can go). I've opened #1260 on top of your branch addressing those, with Windows-tagged tests that exercise a real `cmd.exe → ping.exe` tree. Merging this one now.

PR #1235, opened 2026-09-10
fix(registries): block IPv6 transition addresses in the SSRF guard

aroh3006:
## Description

`isBlockedIP` is the single predicate behind the pre-flight URL check, the dial-time `Control` guard and the app-layer target-host guard. It checks `IsLoopback`/`IsPrivate`/link-local/etc on the resolved address but never looks at the IPv4 address embedded in an IPv6 transition address, so a registry source URL whose host is `[2002:a9fe:a9fe::1]` (6to4) or `[64:ff9b::a9fe:a9fe]` (NAT64 well-known prefix) passes all three layers. On a host with NAT64/6to4 routing that address reaches the embedded IPv4, so `registry add-source` pointed at one of these forms can still get the daemon to fetch from `169.254.169.254` or a loopback service.

This unwraps 6to4, NAT64 (well-known and local-use prefixes), Teredo and the deprecated IPv4-compatible form and re-checks the embedded address. A NAT64 address wrapping a public IPv4 stays reachable. Same class as CVE-2026-73087 (Dozzle).

## Testing
- [x] All existing tests pass

`go test ./internal/registries/` passes. The new transition-address entries in `TestIsBlockedIP` are accepted by the current guard and rejected after the change; the existing entries still pass. `gofmt` and `go vet` clean.

### DIFF
--- a/internal/registries/ssrf.go
+++ b/internal/registries/ssrf.go
@@ -73,13 +73,66 @@ func isBlockedIP(ip net.IP) bool {
 			return true
 		}
 	}
-	return ip.IsLoopback() ||
+	if ip.IsLoopback() ||
 		ip.IsPrivate() ||
 		ip.IsLinkLocalUnicast() ||
 		ip.IsLinkLocalMulticast() ||
 		ip.IsInterfaceLocalMulticast() ||
 		ip.IsMulticast() ||
-		ip.IsUnspecified()
+		ip.IsUnspecified() {
+		return true
+	}
+	// IPv6 transition addresses (6to4, NAT64, Teredo, IPv4-compatible) embed an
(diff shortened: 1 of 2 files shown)


Dumbris:
Thank you @aroh3006 — clean fix, and the CVE-2026-73087 parallel is exactly the class we want closed. Verified the 6to4/NAT64/Teredo/IPv4-compatible offsets against the RFCs and probed the dial-time guard with the allow-private flag forced off; no bypass found. One nit I'll handle in a follow-up: the Teredo vector `2001:0:0:0:0:0:80ff:fffe` has a 0.0.0.0 server field, so it is blocked before the XOR'd client is consulted — swapping in `2001:0:4136:e378::80ff:fffe` pins that branch. Merging.

PR #1319, opened 2026-09-19
fix(shellwrap): Git Bash on Windows cannot exec backslash-style paths

### LINKED ISSUE #1318: Windows: Git Bash cannot exec backslash-style Windows paths, breaks every wrapper-script stdio server
## Summary

On Windows, when the login shell resolves to Git Bash / MSYS (`$SHELL` set, or otherwise detected as bash-like), `shellwrap.WrapWithUserShell` correctly single-quotes a backslash-style Windows path (e.g. `C:\ProgramData\foo\bar.cmd`), but Git Bash still cannot execute it. MSYS's exec layer only resolves POSIX-style paths (`C:/ProgramData/foo/bar.cmd`); a backslash-style path falls through to bash's `PATH` lookup, which fails with `command not found` and — in its own error rendering — drops the backslashes entirely, producing a confusing message like:

```
/usr/bin/bash: line 1: C:ProgramDataQalatCybermcpproxy-wrappersnexusmods-mcp.cmd: command not found
```

This affects **every** stdio server whose `command` is an absolute Windows path (wrapper `.cmd`/`.bat` scripts are the common case) on any host where the resolved login shell is bash-like — i.e. any Windows box with Git Bash and `$SHELL` set, which is common for anyone using Git Bash / VS Code's integrated terminal / MSYS2 as their default shell.

## Repro

1. On Windows, with `$SHELL` set to a Git Bash path (e.g. `C:\Program Files\Git\bin\bash.exe`), configure a stdio server with an absolute Windows-path command, e.g.:
   ```json
   { "command": "C:\ProgramData\SomeVendor\wrappers\foo.cmd" }
   ```
2. Start mcpproxy and watch it try to connect. It fails every retry with:
   ```
   failed to connect: stdio transport (command="C:\Program Files\Git\bin\bash.exe", ...):
   server process exited before completing the MCP initialize handshake; recent stderr:
   /usr/bin/bash: line 1: C:ProgramDataSomeVendorwrappersfoo.cmd: command not found
   ```
3. Confirmed with a minimal Go reproduction calling `shellwrap.WrapWithUserShell` directly and then `exec.Command(shell, args...)` on the result — the quoting is correct (backslashes are preserved inside the single-quoted string), but the resulting `bash -l -c '...'` invocation still fails to exec, proving the bug is in MSYS's own path resolution, not in the Go-side quoting.

## Root cause

`internal/shellwrap/shellwrap.go`, `WrapWithUserShell`: when the resolved shell is bash-like (`isBashLikeShell(shell)` is true) and `runtime.GOOS == "windows"`, the command and args are still shell-escaped and joined verbatim, backslashes and all. MSYS/Git-Bash's `exec()` implementation does not translate a backslash-style absolute Windows path to POSIX form before attempting to run it, so the exec attempt silently fails and falls through to a `PATH`-name lookup, which is what produces the garbled `command not found` message.

## Fix

Convert backslashes to forward slashes in the command and each argument *before* shell-escaping, but only on the Windows + bash-like-shell branch (Windows itself accepts forward-slash paths interchangeably with backslash ones via `CreateProcess`, and MSYS accepts them natively, so this is safe in both directions).

```go
// Git Bash / MSYS on Windows cannot exec a backslash-style Windows path
// (C:\ProgramData\...) even when it is correctly single-quoted: MSYS's
// own exec layer only resolves POSIX-style paths, so it falls through to
// bash's PATH lookup, which reports "command not found" using its own
// mangled rendering of the argv word (backslashes silently dropped). A
// forward-slash path (C:/ProgramData/...) is accepted by both Windows'
// CreateProcess and MSYS's exec layer, so convert before quoting when
// we're about to run a bash-like shell on Windows.
toBashPath := func(s string) string {
    if runtime.GOOS == osWindows && isBash {
        return strings.ReplaceAll(s, `\`, "/")
    }
    return s
}
```

Verified locally:
- Unit test reproducing the original bug (single-quoted backslash path still fails to exec through real Git Bash — `exit status 127`, `command not found`, backslashes dropped from bash's own error message).
- Unit test proving the fix (`bash -l -c '<forward-slash form>'` successfully execs a real Windows binary, e.g. `whoami.exe`, through the wrapped path).
- Full `internal/shellwrap` and `internal/upstream/...` test suites pass after the change; one existing Docker-isolation test (`TestSetupDockerIsolationShellWrapsWhenDaemonEnvMissingNonDarwin`) needed its assertion updated to accept either separator style, since the Docker shell-wrap fallback path goes through the same code and was hitting the identical latent bug.
- End-to-end: rebuilt mcpproxy with the fix and reconnected a real config with ~14 wrapper-`.cmd`-based stdio servers that were all failing with this exact error — all connected successfully afterward (connected server count went from 13 to 27, tool count from 249 to 403).

Happy to open a PR with this change (plus the updated test) if useful — flagging as an issue first in case there's a preferred approach or an existing fix already in flight.

## Environment

- OS: Windows 11 Pro
- mcpproxy-go: v0.67.0
- Shell: Git Bash (`$SHELL` set to Git's `bash.exe`)

LocoLoboZ:
## Summary
- Fixes #1318: on Windows, when the resolved login shell is Git Bash / MSYS, `WrapWithUserShell` correctly single-quoted an absolute Windows path but Git Bash still couldn't execute it — MSYS's exec layer only resolves POSIX-style paths, so a backslash path falls through to bash's `PATH` lookup and fails with `command not found` (mangling the path in its own error output).
- This broke every stdio server configured with an absolute Windows-path command (wrapper `.cmd`/`.bat` scripts are the common case) whenever `$SHELL` resolves to a bash-like shell on Windows — a common setup for anyone using Git Bash / MSYS2 / VS Code's integrated terminal as their default shell.
- Fix: convert backslashes to forward slashes in the command and args before shell-escaping, but only on the Windows + bash-like-shell branch. Both `CreateProcess` and MSYS's exec layer accept forward-slash paths.
- Updated `TestSetupDockerIsolationShellWrapsWhenDaemonEnvMissingNonDarwin`, which exercises the same shell-wrap code path for the Docker isolation fallback and was hitting the identical latent bug — its assertion now accepts either separator style.
- Added `TestWrapWithUserShell_GitBashCanExecWindowsPath`, a Windows-only regression test that actually execs a real binary (`whoami.exe`) through the wrapped Git Bash command, so a regression here fails loudly instead of only failing a string-quoting assertion.

## Test plan
- [x] `go test ./internal/shellwrap/...` — pass, including the new regression test
- [x] `go test ./internal/upstream/...` — pass (all subpackages)
- [x] `go build ./...` — clean
- [x] `gofmt -l` on changed files — clean
- [x] `go vet ./internal/shellwrap/... ./internal/upstream/core/...` — clean
- [x] Manual end-to-end: rebuilt `mcpproxy` with this change and reconnected a real config with 14 wrapper-`.cmd`-based stdio servers that were all failing with this exact error before the fix. All connected afterward — connected server count went from 13 to 27, tool count from 249 to 403.


Dumbris (changes requested):
Thanks for the detailed writeup and the manual verification with 14 wrapper-`.cmd` servers — that's a great regression to have caught and a clear repro. The fix direction (stop keying the quoting dialect on `GOOS` and instead ask "what shell will actually parse this") is the right one.

I did find a few things worth another pass before merging, mostly around how broadly the new backslash→forward-slash conversion applies:

1. **`toBashPath` rewrites backslashes in *every* arg, not just paths** ([`internal/shellwrap/shellwrap.go`](internal/shellwrap/shellwrap.go) around the new closure). It's applied unconditionally to `command` and every element of `args` whenever GOOS=windows and the shell is bash-like, with no check that the string is actually a Windows path. That'll also mangle non-path values that happen to contain a literal `\` — e.g. a SQL Server instance name like `.\SQLEXPRESS`, a `DOMAIN\user` credential, a regex (`\d+\.log`), or a Docker `-e KEY=VALUE` value injected via `injectEnvVarsIntoDockerArgs` — silently, with no error. Might be worth scoping the rewrite to things that look like Windows paths (e.g. `^[A-Za-z]:\\` or UNC `\\`) rather than any string containing `\`.

2. **The new regression test may not exercise the fix it's meant to guard.** `TestWrapWithUserShell_GitBashCanExecWindowsPath` never sets `$SHELL` before calling `WrapWithUserShell`, and `resolveLoginShell()` reads `$SHELL` first — so on a Windows box/runner where `$SHELL` isn't set, it silently falls through to the `cmd.exe` branch instead of the Git Bash one, and the test would pass without ever touching the code path it's testing. On top of that, `.github/workflows/unit-tests.yml` only runs `windows-latest` on push, not on `pull_request` — so as written this test can't gate a PR either way. Could you add `t.Setenv("SHELL", bashPath)` and see if the Windows job can run on PRs (or is skipped intentionally for cost reasons)?

3. **Smaller / lower-confidence, worth a look:**
   - `isBashLikeShell`'s `strings.Contains(lower, "sh")` also matches `pwsh.exe`, `fish`, `csh`/`tcsh`. It previously only picked argv flags (`-l -c` vs `/c`); now it also gates the quoting dialect and the backslash rewrite, so a false-positive there (e.g. `$SHELL=pwsh.exe`) would now silently mis-quote/mis-convert rather than just fail loudly.
   - `insertCidfileIntoShellDockerCommand`'s literal `"docker run"` substring match (`internal/upstream/core/connection_docker.go`) breaks once the docker path is quoted — which happens for the very common Docker Desktop default install path `C:\Program Files\Docker\...`. Pre-existing, but directly hit by this PR's target scenario, so flagging it here.
   - The public `Shellescape()` still keys its dialect on bare `runtime.GOOS`, unaware of `isBash` — its doc comment ("mirrors... so both code paths can converge") is now a bit stale, and a future caller of it could reintroduce this exact bug.

None of this is a knock on the diagnosis or the manual verification — the root cause and the general direction look right to me. Requesting changes mainly for #1 (the unscoped rewrite) and #2 (test coverage), since those affect correctness/confidence in the fix itself. Happy to take another look once addressed!

